Issue #19, opened 2017-11-21
typo / wrong import in digicampipe/utils/geometry.py


dneise:
these two lines don't work:

-from ctapipe.instrument.camera import CameraGeometry
-from ctapipe.instrument.camera import _find_neighbor_pixels as find_neighbor_pixels

should be:

+from digicampipe.instrument.camera import CameraGeometry
+from digicampipe.instrument.camera import _find_neighbor_pixels as find_neighbor_pixels

in order to make it work as explained here:
[CrabAnalysis.pdf](https://github.com/calispac/digicampipe/files/1492710/CrabAnalysis.pdf)

calispac:
This is an old version of installation procedure. see 
[package_installation_Petr_Hamal_20171121.pdf](https://github.com/calispac/digicampipe/files/1492828/package_installation_Petr_Hamal_20171121.pdf)
 Basically ctapipe should be installed from fork in calispac/ctapipe not from cocov/ctapipe anymore

We shall write a correct README for installation.

dneise:
#21 takes up on the problem of the README. So this can be closed

PR #20, opened 2017-11-21
fix typo / wrong import

dneise:
I'd like to merge these changed upstream, so the stuff works as explained in CrabAnalysis.pdf c.f. #19


dneise:
Closed, since this was a non-issue ... simply the documentation I had was wrong.

PR #23, opened 2017-11-22
Install docu in readme

dneise:
Have not tested it yet,

please note these changes, made to the original docu by Petr:

1. we let the user make an **conda env** but we do not use it for the protoZfitsreader. 
    We can however, let the user install it to the env, all they have to do, is use a different path in their LD_LIBRARY_PATH
    I explained how one finds out that path. I tested this, it works.
    **However** this part of the installation is by far the worst. This should be made pip installable, so at least the LD_LIB_PATH is set automatically in the correct way.

2. I added `ipython` to the list of packages to be installed. It is just good.

3. I propose to make the ctasoft folder early and cd into is, so that the zfits stuff is not inside the home, but nicely tidied up.

4. I removed a lot of "ch into, then python setup.py install, then cd out* in the bottom.

5. I propose to install all packages in "editable" mode, so one can edit them in place and see the effect without the need to uninstall-reinstall. (This does not work with C/C++-extensions afaik, they need to be recompiled)

Please have a look.

### DIFF
--- a/digicampipe/utils/geometry.py
+++ b/digicampipe/utils/geometry.py
@@ -1,9 +1,8 @@
 from cts_core import camera as cam
 import numpy as np
 import astropy.units as u
-from ctapipe.instrument.camera import CameraGeometry
-from ctapipe.instrument.camera import _find_neighbor_pixels as find_neighbor_pixels
-
+from digicampipe.instrument.camera import CameraGeometry
+from digicampipe.instrument.camera import _find_neighbor_pixels as find_neighbor_pixels
 
 def find_pixel_positions(camera_config_file, source_x=0.*u.mm, source_y=0.*u.mm):
     camera = cam.Camera(_config_file=camera_config_file)
--- a/README.md
+++ b/README.md
@@ -1,2 +1,81 @@
 # digicampipe
 DigiCam pipeline based on ctapipe
+
+# Installation
+
+## Anaconda 
+
+Follow [the anaconda installation instructions](https://conda.io/docs/user-guide/install/linux.html).
+We propose to use the most recent version.
+
+    wget https://repo.continuum.io/archive/Anaconda3-5.0.0.1-Linux-x86_64.sh
+    bash ./Anaconda3-5.0.0.1-Linux-x86_64.sh
+    
+## Create a virtual environment
+
+    conda create -n digicampipe python=3.5
+    source activate digicampipe
+    
+## install necessary libraries and packages for CTA software
+
+    conda install protobuf=3.0.0 numpy six scipy astropy ipython_genutils decorator 
+    conda install matplotlib llvmlite hdf5 ipython
+
+## Prepare a folder
+
+    mkdir ctasoft
+    cd ctasoft
+
+## Get ProtoZFitsReader
+
+This step involves a bit of manual work, but we are working on streamlining it.
+
+    wget www.isdc.unige.ch/~lyard/repo/ProtoZFitsReader-0.41.Python3.5.Linux.x86_64.tar.gz
+    pip install ProtoZFitsReader-0.41.Python3.5.Linux.x86_64.tar.gz
+    
+type
+
+    python -m site
+
+you'll get something similar to this:
+
+    (digicampipe) dneise@lair:~/ctasoft$ python -m site
+    sys.path = [
+        '/home/dneise/ctasoft',
+        '/home/dneise/anaconda36/envs/digicampipe/lib/python35.zip',
+        '/home/dneise/anaconda36/envs/digicampipe/lib/python3.5',
+        '/home/dneise/anaconda36/envs/digicampipe/lib/python3.5/plat-linux',
+        '/home/dneise/anaconda36/envs/digicampipe/lib/python3.5/lib-dynload',
+        '/home/dneise/.local/lib/python3.5/site-packages',
+        '/home/dneise/anaconda36/envs/digicampipe/lib/python3.5/site-packages',    <----- your envs site-packages
+    ]
+    USER_BASE: '/home/dneise/.local' (exists)
+    USER_SITE: '/home/dneise/.local/lib/python3.5/site-packages' (exists)
+    ENABLE_USER_SITE: True
+    
+Note the `site-packages` folder of your environment. Add this folder to your LD_LIBRARY_PATH:
+
+    export LD_LIBRARY_PATH=$HOME/anaconda36/envs/digicampipe/lib/python3.5/site-packages:$LD_LIBRARY_PATH
+    
+
+## Get a bunch of repos (and install in "editable" mode)
+
+    git clone https://github.com/cta-observatory/pyhessio
+    pip install -e pyhessio
+   
+    git clone https://github.com/calispac/ctapipe
+    pip install -e ctapipe
+    
+    git clone https://github.com/calispac/ctapipe-extra
+    pip install -e ctapipe-extra
+    
+    git clone https://github.com/calispac/digicamviewer
+    pip install -e digicamviewer
+    
+    git clone https://github.com/calispac/digicampipe
+    pip install -e digicampipe
+    
+
+You can only run the software, if you have access to example input data. You have to ask somebody for this.
+
+    ipython digicampipe/pipeline_crab.py


dneise:
As I said, I have not yet tested if this works for me ... I am testing it now.

dneise:
We say we need python 3.5 in 
    conda create -n digicampipe python=3.5

I just tried it with py3.6, but if fails: conda says protobuf 3.0.0 needs py 2.7 (???)
    
    UnsatisfiableError: The following specifications were found to be in conflict:
      - protobuf 3.0.0* -> python 2.7*
      - python 3.6*

I am wondering py it works with py3.5. However I was wondering why we insist on this old version of protobuf for software for something that is not even yet built? Should we not try to use the most recent versions?

dneise:
We say we have these dependencies:

    conda install protobuf=3.0.0 numpy six scipy astropy ipython_genutils decorator 
    conda install matplotlib llvmlite hdf5 ipython

But actually we should track dependencies in the setup.py files and let pip handle them.
The less "manual" work is included in the installation, the more it becomes testable, so we can make sure the stuff installs on different architectures, e.g. using travis or so.

dneise:
After this installation, and trying `pipeline_crab.py` I get:

~~~
(digicampipe2) dneise@lair:~/ctasoft2$ ipython digicampipe/pipeline_crab.py 
---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
~/ctasoft2/digicampipe/pipeline_crab.py in <module>()
      1 from digicampipe.calib.camera import filter, r1, random_triggers, dl0, dl2, dl1
----> 2 from digicampipe.io.event_stream import event_stream
      3 from digicampipe.utils import geometry
      4 from cts_core.camera import Camera
      5 from digicampipe.io.save_hillas import save_hillas_parameters_in_text, save_hillas_parameters

~/ctasoft2/digicampipe/digicampipe/io/event_stream.py in <module>()
----> 1 from digicampipe.io import zfits, hdf5
      2 
      3 
      4 def event_stream(file_list, camera_geometry, camera, expert_mode=False, max_events=None, mc=False):
      5 

~/ctasoft2/digicampipe/digicampipe/io/hdf5.py in <module>()
      1 from digicampipe.io.containers import DataContainer
      2 import digicampipe.utils as utils
----> 3 import h5py
      4 
      5 __all__ = ['digicamtoy_event_source']

ImportError: No module named 'h5py'
(digicampipe2) dneise@lair:~/ctasoft2$ conda install h5py
Fetching package metadata ...........
Solving package specifications: .
~~~

This requirement is missing in the setup.py. I would fix it now immediately but I have to fork it first ...takes time... have no time for it....

dneise:
After installing h5py

I get
~~~
digicampipe2) dneise@lair:~/ctasoft2$ ipython digicampipe/pipeline_crab.py 
---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
~/ctasoft2/digicampipe/pipeline_crab.py in <module>()
      1 from digicampipe.calib.camera import filter, r1, random_triggers, dl0, dl2, dl1
      2 from digicampipe.io.event_stream import event_stream
----> 3 from digicampipe.utils import geometry
      4 from cts_core.camera import Camera
      5 from digicampipe.io.save_hillas import save_hillas_parameters_in_text, save_hillas_parameters

~/ctasoft2/digicampipe/digicampipe/utils/geometry.py in <module>()
----> 1 from cts_core import camera as cam
      2 import numpy as np
      3 import astropy.units as u
      4 from ctapipe.instrument.camera import CameraGeometry
      5 from ctapipe.instrument.camera import _find_neighbor_pixels as find_neighbor_pixels

ImportError: No module named 'cts_core'
~~~

dneise:
And indeed ... cts_core does not exist. It is part of CTS, which is not to be installed according to the PDF of Petr. ...

Unfortunately I have not more time to test this.

Testing installation instructions involves a lot of uninstallation, which needs to be done carefully, otherwise, stuff stays installed, and thus can go missing from the instructions. So I am sorry, but for now. 

These instructions are clearly not yet ready to be merged.

If somebody now wants to take over and add the part I was missing, they have to unfortunately fork from my personal fork of this, and make a PR into my stuff, which I can then merge in order to finally merge a working version upstream.  In an organisation, people could just add this and push directly into my branch. That would be so awesome.

digicampipe/utils/geometry.py
```diff
@@ -1,9 +1,8 @@
 from cts_core import camera as cam
 import numpy as np
 import astropy.units as u
-from ctapipe.instrument.camera import CameraGeometry
-from ctapipe.instrument.camera import _find_neighbor_pixels as find_neighbor_pixels
-
+from digicampipe.instrument.camera import CameraGeometry
```
dneise: I just see this was a mistake on my side... should not haven ended up in this PR. 
Will remove it soon ... at the moment I am traveling and only have a phone, so in order not to forget this ... just a comment.

Etienne12345:
I haven't really followed all the previous discussion that obviously occurred before this thread. My two (four) cents:
- I find your statement about the protozfits reader being "by far the worst" a bit unprofessional. It is indeed not perfect yet, but as you still do no have a running pipeline, it seems to me that other things are actually worse... Also using this kind of language certainly does not help to collaborate. We both went through the FACT experience, let's not repeat the same mistakes. 
- All dependencies are already listed in the setup.py of the protozfits package. Why it did not install them automatically for you I do not know: it did it correctly on all test setups that I had tried. 
- Generally, it sounds like a bad idea of having various README: we have already 3: the one that I had distributed a while ago, Cyril's one from the repo, and now Petr one. Most probably only the one from the repo should survive (and be up to date ;)
- the protozfits reader uses a pip installer because I have no idea nor time to learn how to make a conda package. It seems to me that we are diverging further and further from ctapipe. Instead of fixing pip installers to match your favourite stuff, I'd rather get rid of pip entirely and move everything to conda instead. Otherwise merging with ctapipe will soon become completely impossible.

dneise:
> I find your statement about the protozfits reader being "by far the worst" a bit unprofessional. It is indeed not perfect yet, but as you still do no have a running pipeline, it seems to me that other things are actually worse... Also using this kind of language certainly does not help to collaborate. We both went through the FACT experience, let's not repeat the same mistakes.

@Etienne12345 you are totally right, what I said there was uncalled for. 
I apologize.

dneise:
> Generally, it sounds like a bad idea of having various README: we have already 3: the one that I had distributed a while ago, Cyril's one from the repo, and now Petr one. Most probably only the one from the repo should survive (and be up to date ;)

I totally agree here. This PR is not about a **new** README, but about filling the emtpy README of this repo.

This repo here belongs to @calispac, so I assume you @Etienne12345  mean this README when you said: "Cyril's one from the repo", right?

I think one README should be enough, or say one documentation ... could consist of multiple pages in the end, right? 

However one question I still have:
The documentations I got, referred to these repos:

 * https://github.com/cta-observatory/pyhessio 
 * https://github.com/calispac/ctapipe
 * https://github.com/calispac/ctapipe-extra
 * https://github.com/calispac/digicamviewer
 * https://github.com/calispac/digicampipe
 * https://github.com/cocov/CTS
 
Now, when you say, and I agree, we should not have multiple READMEs ... what is the right place to provide, this README? 

I think we kind of naturally went for the `digicam` repo ... but was this correct? @calispac  what do you think?

dneise:
> the protozfits reader uses a pip installer because I have no idea nor time to learn how to make a conda package. It seems to me that we are diverging further and further from ctapipe. Instead of fixing pip installers to match your favourite stuff, I'd rather get rid of pip entirely and move everything to conda instead. Otherwise merging with ctapipe will soon become completely impossible.

Here, I feel a little bit unsure where this is going.... 

I did not complain about using pip. I think using pip is good. I also have no clue about making conda packages. 

All the issue I opened was about, and all this PR is about is:

I got docu in order to install this, and it did not work. I assume it was outdated or something was missed. Does not matter errors happen. 

I did not intend to change anything here in the code base, just add docu on how to install it. 
(and in a next PR, test docu, to make sure it works, in order to prevent it from outdating).

If pip or conda is used, I don't mind as long as it works and I can test it.

Etienne12345:
Very quickly:
- Comment: no worries, I've seem worse :)
- readme: yes, digicampipe sounds like the place where to put it. Eventually all these separate packages should be merged to ctapipe
- pip vs conda: I like pip too, but ctapipe uses conda instead. My point was to "keep things messy for now, and only fix/improve if it goes in the ctapipe direction" 

Cheers !

calispac:
@dneise Actually I think we should have a README per packages and the setup.py should resolve dependencies by itself. In the main repr. (cta-sst-1m for instance) a full README or some script that installs everything for new comers. 
But as you have probably seen each repr. is mostly undocumented and for instance the CTS package does not install well i.e. you cannot do : 
`from CTS import cts_core`
you rather have to do
`import cts_core` 

I guess there is a lot of work on the design of each different packages. The way I see it is that we should separate packages when it is possible. For instance CTS was intended to control the calibration setup. But it includes camera hardware mapping which turns out to be useful for digicampipe. So in this case I think we should have a package for the camera mapping and a package for the calibration setup control.

pip or conda (although I don't know what it implies for install) is fine for me as long as we can install the software with `python setup.py install`

dneise:
This is replaced by #25

PR #25, opened 2017-11-22
Add installation docu into README

dneise:
This PR replaces #23. #23 came from a fork, which was unwieldy for me to work with.

The README is based on two PDFs I got and comments from @Etienne12345 and @calispac.

I made some adjustments, to streamline a bit:

1. use $CONDA_PREFIX for the LD_LIBRARY_PATH
2. I added `ipython` to the list of packages to be installed. It is just good.
3. I propose to make the ctasoft folder early and cd into is, so everything is in the same folder
4. I propose to install all packages in "editable" mode, so one can edit them in place and see the effect without the need to uninstall-reinstall. (This does not work with C/C++-extensions afaik, they need to be recompiled)

Please have a look.

---- 
Test:
   It works like this on my machine.

Travis ongoing...

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
 build/
 __pycache__
 .idea/*
-test*
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,105 @@
-build/
-__pycache__
 .idea/*
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*.py[cod]
+*$py.class
+
+# C extensions
+*.so
+
+# Distribution / packaging
+.Python
+build/
+develop-eggs/
+dist/
+downloads/
+eggs/
+.eggs/
+lib/
+lib64/
+parts/
+sdist/
+var/
+wheels/
+*.egg-info/
+.installed.cfg
+*.egg
+MANIFEST
+
+# PyInstaller
+#  Usually these files are written by a python script from a template
+#  before PyInstaller builds the exe, so as to inject date/other infos into it.
+*.manifest
+*.spec
+
+# Installer logs
+pip-log.txt
+pip-delete-this-directory.txt
+
+# Unit test / coverage reports
+htmlcov/
+.tox/
+.coverage
+.coverage.*
+.cache
+nosetests.xml
+coverage.xml
+*.cover
+.hypothesis/
+
+# Translations
+*.mo
+*.pot
+
+# Django stuff:
+*.log
+.static_storage/
+.media/
+local_settings.py
+
+# Flask stuff:
+instance/
+.webassets-cache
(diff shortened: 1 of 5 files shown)


dneise:
I learned by installing it on a travis node, that one actually needs to install zeromq.
I installed it via `apt`, since the travis-node uses Ubuntu. 

I also added this to the README as:

    sudo apt-get install libzmq3 libzmq3-dev

Is that okay like this? Or should it be "conda install"-ed?

calispac:
@dneise  Even if this commands are for some linux distros at least we have it tested on one OS. I guess its fine for now

dneise:
Okay I like this now, @calispac would you merge whenever you are ready?

Etienne12345:
For libzmq.so.3 it seems that it is indeed missing from the pip package of the protozfitsreader.

Instead of stuggling to get it from the system, I could very well add it to the package (as I did for the protobuf libraries). 

Should I go ahead and produce an updated version of the package or do you prefer to continue with the conda installers ?

dneise:
@Etienne12345 
Do you think this question could also be discussed in the protozfitsreader repository? Since it is in principle decoupled from digicampipe?

I personally have no experience with deploying shared objects,  so far I only delivered source code and had the user compile it. I think not having users compile the project, but delivering ready-made-binaries is great, but due to my utter lack of experience with this have absolutely no idea, if there is any downside of this approach.

Regarding this project, I think it is fine at the moment, since one can simply tell people to install `libzmq3`using their package manager, apt, yum, whatever. 

Regarding the version, On my machine I did not install libzmq.so.3 but simply created a symlink from so.5 to so.3 and so far it seems to work. So I was wondering how strict is the version here? Does it absolutely have to be 3? If it also could be 4 (which happens to be available via conda)

Etienne12345:
Oops sorry if this was not the correct place to discuss this. 

It would be difficult to recompile the reader from scratch, because you also get dependencies to protobuf and zmq: changes nothing for the end-user. 

As for the binaries, they -afaik- only use very basic features of the linux kernel, so in principle it should be not problem to distribute them: I expect compatibility with most (if not all) linux distros. 

Your solution of linking the wrong version of the library could also work, but this time we rely on the features of the library rather than the compatibility with the kernel. 

I will add the libzmq.so.3 lib to the package, sound easier than asking the user to do crazy symlinking by him/herself

Etienne12345:
Actually, thinking about this, there is absolutely 0 problem with using any of the zmq libraries. Why you may ask ? Because it is a dependency of the C++ project used to build the bindings, but in reality this lib is not used at all by the reader... I should remove the dependency entirely, but I must rework the whole build system: will take time.

PR #42, opened 2017-12-11
No LD_LIBRARY_PATH

dneise:
I played a bit around and got Etiennes tarball contents so that one does not have to set the LD_LIBRARY_PATH explicitely.

For playing around I put the contents of the tarball into a git repo, so you can have a look at it.
the repo is here: https://github.com/dneise/protozfitsreader

As I said, this is basically the same contents as were in the tarball, I just patched 3 shared libs and made 2 or 3 imports relative in 2 py-files.


dneise:
Erm .. wait .. this PR contains commits which should not be in here ... 
[working on it]

dneise:
This PR was dirty:

    It had too many commits which did not belong together

PR #43, opened 2017-12-11
remove the need to set the LD_LIBRARY_PATH

dneise:
I played a bit around and got Etiennes tarball contents so that one does not have to set the LD_LIBRARY_PATH explicitely.

For playing around I put the contents of the tarball into a git repo, so you can have a look at it.
the repo is here: https://github.com/dneise/protozfitsreader

As I said, this is basically the same contents as were in the tarball, I just patched 3 shared libs and made 2 or 3 imports relative in 2 py-files.


dneise:
So the only "problem" with this installation which I still see is the Python 3.5 dependency, since ctapipe tests on python 3.5 and 3.6 

I have still problems building the C++ sources, so I cannot do this myself, but I would *assume* if Etienne can provide in addition to the file:

    rawzfitsreader.cpython-35m-x86_64-linux-gnu.so

another file, which was compiled against a Python 3.6 and is therefore named:

    rawzfitsreader.cpython-36m-x86_64-linux-gnu.so

Then this remaining issue could be solved.

dneise:
Ah I touched one line in `protozfitsreader.py` and now the pep8bot makes me responsible for it not complying to our style guidelines ... nice :-D

Okay so I'll fix pep8 issues now.

calispac:
I merge this but we should come with a solution for compiling (if there is a good one :D ) and test against python3.6

Etienne12345:
Nice that the LD_LIBRARY_PATH is not needed any longer !

In the past I was making a packge for each python version. I stopped making it for python 3.6 because as you may recall (or not), I experienced very weird errors with Digicampipe on python 3.6, which were resolved by rolling back to python 3.5. It was not even errors, only the pipeline was not producing the expected output. No idea why. 

In addition, as long as we have to leave with a digicampipe separate from ctapipe, I'd stick to one single python version and let it to ctapipe to maintain several version?

calispac:
From what I understood in the issues discussion of ctapipe they are not going to maintain so called "unofficial" formats so we should rather find the solution ourselves for 3.6. Then I doubt that before the "official" format is decided ctapipe will move to an other python. 
I have also no idea why it could not work on 3.6. I would assume that python 3.5 is compatible with 3.6. I will try a fresh install on 3.6

Etienne12345:
I would ask what are the benefits of moving to python 3.6 ? If indeed both versions should be compatible, then we do not care to stick to 3.5 or not. Given the limited manpower, I'd avoid at all costs to maintain more than one version. If you really wanna spend time looking into versionning, I can make the 3.6 reader package: let me know.

dneise:
> I stopped making it for python 3.6 because as you may recall (or not), I experienced very weird errors with Digicampipe on python 3.6, which were resolved by rolling back to python 3.5. It was not even errors, only the pipeline was not producing the expected output. No idea why.

I see, this is new to me, thanks for the information.

dneise:
> I would ask what are the benefits of moving to python 3.6 ? 

Here are the new features in py 3.6. over py3.5 ... dicts are 20..25% smaller in memory, awesome if you ask me :-D

Generally I have the impression always supporting the most recent versions saves one from the greater pain which is sticking to one fixed version and then never being able to support python 7.2 in 5 years, since it is too much work.

I think this matter is solved easiest by having the end-user (or travis for that matter) compile the sources against their version of python. 

I am playing around with it at the moment, but there is one question I will not be able to solve, which is this privacy issue. 

If we cannot deliver the sources to the end-user, since we have to keep the project, closed source, then this "solution" I am currently playing with is of course a non-solution.

Etienne12345:
I am strongly against letting the user recompile all sources. Not because of privacy reasons (reader code is mostly public), but because of user-support reasons. Creating two packages (linux+osx) per python version is trivial (only if we stick to a single supported python of course). Supporting the users who will have the most exotic issues while trying to compile the code: no. If they want to do it anyway, they are of course free to do so, but that should not be the default.

dneise:
Cool so there are no privacy issues. That is great news.

dneise:
This issue with python 3.6, could that have been related to issue #44

Issue #44, opened 2017-12-11
possible issue with `io/protozfitsreader.py`


dneise:
Hello, have a question about a line in `io.protozfitsreader.py`. 

https://github.com/calispac/digicampipe/blob/b5e17ddd6e402710dfdb978c278742ed35f2623c/digicampipe/io/protozfitsreader.py#L324-L330

This question is about line 328, but I am not quite sure I understand the inputs `pixels`, `baselines`.
I assume:
 - `pixels`: a 1D array of (integer) pixel ids, no necessarily ordered, but no duplicates
 - `baselines`: a 1D array of floats, having the same length as `pixels`.

Are these assumptions correct? That is my first question.

-----

Now, first I assumed `pixels` is a **sorted** list of all pixel ids, and thus I did not understand why line 328 is needed at all, since *if pixels is sorted and has no duplicates* then *properties is identical with baselines*.

So I thought: "Ahaaa properties is a sorted version of baselines" :-D
But wait! This is using an implementation detail of dicts, in the C-implementation of python.
(Also it is slow)

Let me show you this is a problem:

I prepared a little test script: `properties.py` with this contents:
```python
import numpy as np

np.random.seed(0)
n_pixels = 5
pixels = np.arange(n_pixels)
np.random.shuffle(pixels)
baselines = pixels + 0.1

print('pixels:', pixels)
print('baselines:', baselines)

properties = np.array(list(dict(zip(pixels, baselines)).values()))
print('properties', properties)
```

I set the random seed, so it executes the same on all our machines:
This is the output one would expect, right?
```
pixels: [2 0 1 3 4]
baselines: [ 2.1  0.1  1.1  3.1  4.1]
properties [ 0.1  1.1  2.1  3.1  4.1]
```

pixels is some list of indices, not sorted, but no duplicates,  baselines is just some floats, but I made them, so we can see which baseline belongs to which pixel. properties now, is a sorted version of baseline. Keep in mind it is not sorted by size, it is sorted according to the order of `pixels`.

Now let me show you how it looks in a py3.5 and a py3.6 environment.

### python 3.5
```
(foo) dneise@lair:~/ctasoft/digicampipe/digicampipe/io$ ipython
Python 3.5.4 |Anaconda, Inc.| (default, Nov 20 2017, 18:44:38) 
Type 'copyright', 'credits' or 'license' for more information
IPython 6.2.1 -- An enhanced Interactive Python. Type '?' for help.

In [1]: %run properties.py
[...]

PR #45, opened 2017-12-11
work on protozfitsreader

dneise:
# Work in progress, do not merge

I open this PR so people can have a look already now. Tests most likely do not pass.
I started this as a refactoring, so change style not function, nor interfaces. 
But while reading, I think I spotted a bug, c.f. #44.

Remarks or Questions are very welcome.


dneise:
I totally agree. Actually I will write the tests first .. so there will be a new PR ... with a couple of tests on the "old" version of this.

dneise:
The we can first look at the tests and see if trust them ... and then we can apply the tests on this new version and *home* the new version still passes them all.

dneise:
Now this branch is passing the same tests, as the current master. 

Of course, since I changed the interface, also the tests needed to be changed: 
A lot of these:
```
zfits = ZFile(example_file_path)
stuff = [zfits.get_stuff() for __ in zfits.move_to_next_event()]
```
were changed into these:
```
stuff = [event.stuff for event in ZFile(example_file_path)]
```

dneise:
BUT: there is one value, which is now -1, and was 0 before:
`num_channels` or `num_gains`. 

I have no idea if zero or minus-one is correct. Strange are both, I would have expected 1.

dneise:
The test for `num_gains` I have marked as "expected to fail" since I have no idea which value should come out of that

calispac:
`num_gains` is the number of amplification channels per pixel. Some cameras have low and high gain. we only have 1. So num_gains should be 1

calispac:
> BUT: there is one value, which is now -1, and was 0 before:
num_channels or num_gains.

I don't know what is the diff. between the two

dneise:
Ha! found it.

So this is how it is in the current master:

    data.inst.num_channels[tel_id] = zfits.event.num_gains

so something called `num_gains` is assigned to something called `num_channels`. **and**  `zfits` has this

    class Zfile:
        def get_num_channels(self):
            return self._get_numpyfield(self.event.head.numGainChannels)

Here when one calls `get_num_channels` one actually gets `numGainChannels`

-----

This confused me and when refactoring the protozfitsreader, i.e. `ZFile` I **wrongly** assumed `get_num_channels` would be used like this:

    data.inst.num_channels[tel_id] = zfits.get_num_channels()

But it is not! It is not used **at all**.

----

Solved this now... The value is still 0 and not 1 .. in master and in this branch

calispac:
@Etienne12345 Do you know why num_channels is 0 not 1 ?

dneise:
I've added a test, that benchmarks `zfits_event_source` on a 100 event file, with and without `expert_mode==True`. 

Both are more or less equally fast.

dneise:
Ah i think this is ready now, sorry

calispac:
Does this look correct now ?

PR #46, opened 2017-12-11
Fix install

calispac:
The owner of CTS package has changed. the path is corrected.


.gitignore
```diff
@@ -1,3 +1,5 @@
+build/
+__pycache__
```
dneise: isn't this a duplicate of line 5?

.gitignore
```diff
@@ -103,3 +105,6 @@ venv.bak/
 
 # mypy
 .mypy_cache/
+
+# Baobab slurm script
+job.sh
```
dneise: pep8: a file should end with a newline

dneise:
Ah what is `job.sh` and why do we gitignore it?

.gitignore
calispac: it was a file I am using on a cluster machine.

calispac:
a file that was used for a cluster machine

dneise:
Looks pretty fine, I'd say we can merge as soon as travis is done

PR #48, opened 2017-12-11
More tests

dneise:
I added some more tests, which document what kind of stuff we expect to come from the example file.

These tests all pass, since I assume the current reader implementation to be correct.
I have however some questions based on these observations:

### event number
`zfits.get_event_number()` and `zfits.event.eventNumber` contain the same values. Are both needed?
If so, what should be the semantic difference beween these two numbers?

### num_gains
What does this number mean? It is always zero, should it not be 1? 

### event number again
it does neither start a 1 nor at zero. but at a "seemingly" random value. Is this a problem?

### DIFF
--- a/digicampipe/tests/test_read_raw_events.py
+++ b/digicampipe/tests/test_read_raw_events.py
@@ -39,6 +39,20 @@ def test_rawreader_can_work_with_relative_path():
     event.ParseFromString(raw)
 
 
+@pytest.mark.xfail
+def test_rawreader_can_read_runheader():
+    from protozfitsreader import rawzfitsreader
+    from protozfitsreader import L0_pb2
+
+    rawzfitsreader.open(example_file_path + ':RunHeader')
+
+    raw = rawzfitsreader.readEvent()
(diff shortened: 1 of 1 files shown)


dneise:
I think these tests are not entirely useless ... so we should merge them.

digicampipe/tests/test_read_raw_events.py
```diff
@@ -12,7 +13,22 @@
...
+    1.5094154944218719e+18,
+    1.5094154944245553e+18,
+    1.5094154944267853e+18,
+    1.509415494438982e+18,
+    1.5094154944452902e+18
+]
+EXPECTED_GPS_TIME = [0] * EVENTS_IN_EXAMPLE_FILE
 
 @pytest.mark.skip(reason="we know the current version does not raise")
 def test_zfile_raises_on_wrong_path():
```
calispac: should it be : def test_zfile_n_rows() ?

calispac: ah I did not quiet understand...

PR #49, opened 2017-12-12
Easier installation

dneise:
Okay, I like this.  Installation is drastically easier now.

A fresh installation on a travis machine (which might not be the fasted machines in the world, since they are virtual boxes) takes 5 minutes.

The tests (in the master there are only 6 tests at the moment) take ~2 seconds. 

So everybody should now be able to build this is less than 10 minutes on their own box. At least I hope so.


calispac:
Yes definitely easier this way ! thanks

PR #50, opened 2017-12-12
removed .idea/ folder

calispac:
this just to remove the folder digicampipe/.idea.


calispac:
ready to merge

dneise:
adding something to the `.gitignore` file does not remove it from the repo.

    git rm path/to/stuff/to/remove*

this removes some stuff.

Adding something to the gitignore should help people to not accidentally add something when doing

    git add some/new/folder

Because, when adding a new folder, everything inside is added. One can avoid that by adding stuff explicitely

    git add some/new/folder/*.py

for example only adds py-files ...

dneise:
Do you still want to keep `digicampipe/.idea/*` in the gitignore?

calispac:
digicampipe/.idea/ is a folder built by PyCharm (my IDE). Is it useful to add it in the .gitignore ?

PR #51, opened 2017-12-12
Lid cdd analysis

calispac:


dneise:
This is a pretty large PR, which usually is hard to review. Smaller PRs touching only 1 or 2 files are often easier to review.

But this PR seems to only **add** features never take away ir modify existing interfaces.

Only in camera.dl2 it looks like this is actualky changing an existing interface.

Since dl2 has no unit tests yet, it is not easy to guess if this PR breaks existing features somewhere else.

dneise:
Ah I forgot, sorry.
This is a very nice PR and the new features are interesting and it show be merged as soon as possible.

Despite it having some pep8 issues, we could also merge it now and tidy it up later. That is perfectly fine as long as it has a low chance of breaking existing features.

calispac:
it will break dl2.py I opened the PR my self but I should have let @yrenier do it

dneise:
Yes thats maybe better ... still it was nice to have a look at the code already.

yrenier:
Sorry it seems there was a problem rebasing to the current master, stuff got included but should not have (like the dl2.py and container.py changes). I'll fix this and ask for the merge myself :-D

Issue #52, opened 2017-12-13
Install on baobab machines does not work correctly


calispac:
I will leave more details later but basically the install on cluster machines at UNIGE showed that the [protozfitsreader](https://github.com/dneise/protozfitsreader) is not installing well with ananconda.

When trying to do 

```python
import protozfitsreader
```

I get 
```python
from protozfitsreader import rawzfitsreader
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/alispach/ctasoft/src/protozfitsreader/protozfitsreader/init.py", line 6, in <module>
    from . import rawzfitsreader
ImportError: /usr/lib64/libstdc++.so.6: version `GLIBCXX_3.4.20' not found (required by /home/alispach/ctasoft/src/protozfitsreader/protozfitsreader/rawzfitsreader.cpython-35m-x86_64-linux-gnu.so)
```

doing 

```bash
ldd src/protozfitsreader/protozfitsreader/rawzfitsreader.cpython-35m-x86_64-linux-gnu.so
```

showed 

```bash
src/protozfitsreader/protozfitsreader/rawzfitsreader.cpython-35m-x86_64-linux-gnu.so: /usr/lib64/libstdc++.so.6: version `GLIBCXX_3.4.20' not found (required by src/protozfitsreader/protozfitsreader/rawzfitsreader.cpython-35m-x86_64-linux-gnu.so)
        linux-vdso.so.1 =>  (0x00007ffcb45e4000)
        libZFitsIO.so => /home/alispach/.local/lib/python3.5/site-packages/libZFitsIO.so (0x00002ad56c92b000)
        libpython3.5m.so.1.0 => not found
        libstdc++.so.6 => /usr/lib64/libstdc++.so.6 (0x00002ad56cbd5000)
        libm.so.6 => /lib64/libm.so.6 (0x00002ad56cedb000)
        libgcc_s.so.1 => /lib64/libgcc_s.so.1 (0x00002ad56d160000)
        libpthread.so.0 => /lib64/libpthread.so.0 (0x00002ad56d376000)
        libc.so.6 => /lib64/libc.so.6 (0x00002ad56d593000)
        libACTLCore.so => /home/alispach/.local/lib/python3.5/site-packages/libACTLCore.so (0x00002ad56d928000)
        libz.so.1 => /lib64/libz.so.1 (0x00002ad56dbba000)
[...]

PR #54, opened 2017-12-13
Install protozfitsreader v0.41.1

dneise:
There was still a problem with the dependencies of "rawzfitsreader.[...].so".
They should now be solved in protozfitsreader v.0.41.1 which is installed after this PR

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,5 +22,6 @@ script:
   - conda list
   - pytest -vv digicampipe
   - ldd /home/travis/build/calispac/digicampipe/src/protozfitsreader/protozfitsreader/rawzfitsreader.cpython-35m-x86_64-linux-gnu.so
+  - python -c "from digicampipe.io.protozfitsreader import ZFile"
   - python -c "import protozfitsreader"
   # - python pipeline_crab.py   ## this needs input files, commented out until test files ready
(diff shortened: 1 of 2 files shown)


dneise:
I think the environment.yml can be reduced in size a bit .. c.f. https://github.com/dneise/protozfitsreader/blob/master/environment.yml

dneise:
I think we should merge this.

The diff is small.

I do two more tests on travis: basically trying to "import protozfitsreader" 

This did not work before...

I had to modify the RPATH inside the rawzfitsreader.[...].so in the protozfitsreader repo ... this was done in v0.41.1

dneise:
This should solve #52

calispac:
Nice that we have releases of protozfitsreader. We can keep track of the new fields added.

dneise:
Well Etienne is already keeping track with his release number.

Actually I destroyed his versioning system, when moving the contents of his tarball into a git repo. and said in our requirements.txt:

    --e git+https://github.com/dneise/protozfitsreader.git#egg=protozfitsreade

Only now I realized that was not very good (actually it was bullshit) ... and I reintroduced what Etienne already had done:

    +https://github.com/dneise/protozfitsreader/archive/v0.41.1.tar.gz

PR #55, opened 2017-12-13
Analysis of lid ccd images.

yrenier:
Allow to find stars and calculate galactic coordinates from them.
Allow to find precise cones position too.


yrenier:
argh an other failure ....

PR #57, opened 2017-12-13
Analysis of lid CCD images

yrenier:
Get cones positions in lid CCD images.
Get pointing from stars reflections on the lid.


yrenier:
I tried to include test for the pointing, but I'm not sure how to handle the dependency on astrometry.net and the indexes files which are quite big

dneise:
Hey Yves. This is a nice PR. pretty big though... pretty hard to review... little PRs are much much easier.

This is mostly adding features, not modifying existing ones.
I see some possibilities to improve style and possibly runtime, also a few more tests might help to explain what this PR is all about.

Would you like me to go through the PR and make a few comments? Or would you rather merge these new features and later tidy up the code. Both is fine for me.

dneise:
Ah I just made one change in the `environment.yml` so the build runs through. And then I modified the test a bit, so it finds the input files.

So tomorrow before working you should not forget to: git pull this branch, since I pushed into it.

dneise:
I'm sorry I just pushed into your branch ... I just wanted to see if the test worked.

digicampipe/tests/test_lid_ccd.py
```diff
@@ -1,6 +1,13 @@
 from digicampipe.image.sky_image import *
 import os
 
+from pkg_resources import resource_filename
```
dneise: Ah maybe I should explain what I did here. So before I made this change to the tests. We got this error on the continous integration machine:
FileNotFoundError: https://travis-ci.org/calispac/digicampipe/builds/316067756#L956

The reason is that this relative path does not work, when the test is not executed from the /tests folder. And it usually is not. 

To find test resources on the customers system, after the package was installed, there is the module called `pkg_resources`.  This combined with `glob` allows us to find all the fits files. 

For the 2nd part ("'tests/resources/stars_on_lid/*.fits'") I should have used `os.path.join` and write it like:

    os.path.join('tests', 'resources', 'stars_on_lid', '*.fits')

So the path is platform independent. Would be a pity to be the reason why this package does not run on Windows10.

But I was lazy :-( ... sorry.

I also pulled the path finding and list generation out of the test function. This is because I like the test function to contain basically only two things: 
    calling the function/object under test
    asserting the result to be correct.

Generation of "difficult" input arguments, is nothing I want to test. I am sure I can make this list of file paths.
Also I want to tell this to the reader: this test is about calling the `LidCCDObservation` constructor and not about making a list of strings. I do not want to distract the reader from the actual thing under test ... but maybe .. this is not so important here...

dneise:
Ah regarding astronetry.net I do not know this project, but they have a web API, maybe one can use that instead of calling the thing locally?

yrenier:
First thanks for fixing the requirements and access to the test files!
I know it is a large PR, but I was not decided until yesterday if this should be part of digicampipe.
About access to the web API, I found a python client (https://github.com/dstndstn/astrometry.net/blob/master/net/client/client.py). I'll get a closer look on it tomorrow, to see if the needed feature are exposed.

dneise:
I also had a brief look at the web-client of astrometry.net 
To me personally it seems that sending images to their server during our tests might be a little unfair, ... 

I did not know astrometry.net before, thanks for showing it to me.

Anyway, regarding CI of digicampipe:
1.) Maybe we just install astrometry.net on the travis node.
If this turns out to be impossible we could say:
2.) We do not want to test that astrometry works correctly, but that we provide the correct **input** for astrometry.net. If I've understood the code correctly, you create a fits file from the cropped image and then call an executable (using subprocess) with this fits file. 
So your part is essentially done when the fits file is created. 
Now one could mock-up the behaviour of astrometry.net and write a mock-astrometry.net, which looks at the fits file and if it contains the correct stuff it returns a **pre-recorded** answer of astrometry.net. 
3.) If both does not work, you one could setup a test server in your office.
Once per day or hour? It checks out the master, builds everything and runs the complete test-suite and sends the result via email to everybody.
On travis we can skip some tests by marking them with for example 'needs-astrometry' 

This is not ideal, but better than nothing.

So I'd say lets first try and install it on travis. Can you point me to a good installation instruction for astronetry.net? All I found looked pretty outdated to me.

yrenier:
I gave basic instructions at the end of the README.md: 
https://github.com/calispac/digicampipe/blob/lid_ccd/README.md

dneise:
Ah nice, so should we try to install it on a travis box during our unit tests? might result in longer running tests, and more power consumption on the travis cluster :-D but as long as we do not get angry emails from them ... :-D

README.md
```diff
@@ -37,3 +37,25 @@ your prompt should look similar to this this:
...
+    cd ~/astrometry-data
+    wget http://broiler.astrometry.net/~dstn/4200/wget.sh
+    chmod +x wget.sh
+    ./wget.sh
+
+Grab a beer as it will take a while ... Also, indexes 4203 and lower are probably not needed.
+
+Then add the folowing line to /etc/astrmetry.cfg:
+
+    add_path ~/astrometry-data
```
dneise: I do not know the commandline tool `add_path`. I assumed this is a part of astrometry.net, so I installed it on my machine ... no luck. I still don't have `add_path`

yrenier: that line should be inserted in the /etc/astrmetry.cfg file (the location whithin that file is probably not important).

dneise: Ah .. I am to stupid to read your instructions ... sorry

dneise: omg that 10GB

dneise: If we let travis download that for every test, the astropy server load will spike :-D

dneise:
Independent of the astrometry discussion, I think it would be maybe good to try and work on this UnboundLocalError here

https://travis-ci.org/calispac/digicampipe/builds/316072426#L989

README.md
```diff
@@ -37,3 +37,25 @@ your prompt should look similar to this this:
...
+
+## Pointing using lid CCD images
+
+To be able to determine the pointing using stars reflections on the lid, you need to install astrometry.net:
+
+    sudo apt-get install astrometry.net
+    
+You will need to download catalogs too: 
+
+    sudo mkdir ~/astrometry-data
```
dneise: I think you do not need to use `sudo` to make directory in your home, unless it is important that this directory belongs to the `root`. Is that important?

README.md
yrenier: Is it possible to put astrometry in the initial image travis is using to build the software ?
On the long term, I think the solution using the web service is the one we should to go. However there is an API key which must be used, which probably means there is an usage cap.

dneise: I think custom images are part of their enterprise features, i.e. we don't get it for free. Not sure is CTA is ready to actually pay for one of the most important practices in software development .....

yrenier:
I'm looking at the UnboundLocalError, but I cannot reproduce it on my computer.
I think the error comes from example_lid_CCD_image_file_paths being empty for the travis build. I'll try rising an exception in that case to see if it is the problem.

README.md
yrenier: I guess then I will have to go for the client connecting to the webservice.

yrenier:
It appears either resource_filename or glob is failing on travis.

dneise:
> It appears either resource_filename or glob is failing on travis.
Almost correct. The installation did not what we expected it to do. 
By default `setup.py` only installs `*.py` files to the installation destination. So test resources, which we want to be delivered also to the end user we must explicitly mention in the setup.py file in `package_data`. 

We forgot to add the new `stars_on_lid` folder and so setup.py did not copy those fits files to the installation directory and so they could not be found by glob in the given directory.

Developers usually do not see this, since they install the package in "--editable" mode, but users see it .. and cannot execute the tests ... and then one ends up with the "but it works on my machine" effect.

dneise:
okay .. as expected .. now the problem is only "solve-field" ...

yrenier:
great, thanks !

dneise:
Do you also plan to add a test that does the hexagon finding FFT stuff, you showed at the last analysis telcon?
Or is ir already there and I just missed it?

yrenier:
good point, I still have to do that.

yrenier:
The test for precise determination of camera pixels position is done and passing :-)

README.md
moderski: I think we should assume that astrometry.net is installed locally, and it is a requirement during installation (not part of the digicam pipe distribution). The digicam pipe should test if the existence of astrometry.net is tested, but not the location of the current package. For tests, a mockup astrometry.net should be produced which accepts images and return astrometry.net output (as suggested by Dominik).

yrenier: Thanks for the input, but I already got the client for web API of astrometry.net working today. It makes things much easier as I can test if astrometry is available locally and, if not, use the web service.

yrenier:
Test about the determination of the pointing based on stars reflections passing.
Tagging of images submitted not working yet, waiting for update of astrometry.net

digicampipe/tests/test_lid_ccd.py
```diff
@@ -0,0 +1,54 @@
+from digicampipe.image.sky_image import *
+from digicampipe.image.cones_image import *
```
dneise: `from foo import *` should be avoided. It clutters global namespace and causes confusion for people who do not know by heart, what symbols the imported modules contain.

pointing_from_stars.py
```diff
@@ -0,0 +1,30 @@
+from digicampipe.image.sky_image import *
```
dneise: `import *` should be avoided

digicampipe/tests/test_lid_ccd.py
```diff
@@ -0,0 +1,54 @@
...
+    crop_pixels1 = [  # (850, 50),
+        # (550, 500),
+        (350, 900),
+        # (600, 1350),
+        (850, 1800), ]
+    crop_pixels2 = [  # (1100, 650),
+        # (900, 1000),
+        (770, 1550),
+        # (1000, 2050),
+        (1300, 2400), ]
```
dneise: a rectangle is defined by the the combination of an element in crop_pixels1 and an element in crop_pixels2, which share the same index. So one could say:

    rectangles = list(zip(crop_pixels1, crop_pixels2))

I think currently the interface is very easy to misunderstand or misuse, one might by accident give a too long list as `crop_pixel1` and so on.

I assume it would be easier to invent a small class `Rectangle`, e.g. like this:

    class Rectangle:
        def __init__(self, bottom, left, top, right):
            # ...

And give `LidCCDObservation` a list of rectangles.

digicampipe/image/sky_image.py
```diff
@@ -0,0 +1,555 @@
...
+        if output_dir is None:
+            plt.show()
+        else:
+            output_filename = self.filename.replace('.fits', '-treated.png')
+            plt.savefig(output_filename, bbox_inches='tight', pad_inches=0)
+            plt.close(fig)
+            print(output_filename, 'saved.')
+
+
+class LidCCDObservation:
```
dneise: why is this a class? It looks like it is actually a function which returns `lidccd_images` the 3 methods this class has, are actually functions, which act on lists lists of ccd_images.

digicampipe/image/sky_image.py
```diff
@@ -0,0 +1,555 @@
...
+        :param threshold: threshold: pixels with a smaller value than threshold after subtraction are set to 0. Default at 100x the
+            average pixel value after subtraction.
+        :param scale_low_images_deg: minimum field of view (in degrees) considered during fitting.
+        :param scale_high_images_deg: maximum field of view (in degrees) considered during fitting.
+        :param guess_ra_dec: aproximate pointing coordinates
+        :param guess_radius: radius of pointing  coordinates around guess_ra_dec tried during determination
+        """
+        self.lidccd_images = []
+        image_shape = None
+        if type(filenames) is not list or len(filenames) == 0:
```
dneise: this check is not actually needed. two lines below this check in:

    for filename in filenames:
    
An exception will be raised, if filenames is an empty list.

It is not really advisable to explicitely check for the type `list` here. Since also a `tuple` of filenames would be okay. Also any iterable would be okay, which spits out filenames, when iterated over like `glob.iglob` for example.

dneise: Explicite type checks are generally considered unpythonic... the mindset is: If it sounds like a duck and looks like a duck, it probably is a duck.
That means, if my code can work with the stuff it was given, then I am fine with it, if not, the code will raise an exception in exactly that place, where the problem occured. There is generally no need for me as a developer to explicitely check a type.

digicampipe/image/utils.py
```diff
@@ -0,0 +1,416 @@
...
+        rounded_factors = np.round(factors[is_used]).reshape(-1, 1)
+        ks2s_rescaled = 1 / rounded_factors * ks[is_used, :]
+        ks_base[1, :] = np.mean(ks2s_rescaled, axis=0)
+        break
+    ks_base[ks_base[:, 0] < 0, :] *= -1
+    return ks_base, auto_correlation_saved, center_peaks
+
+
+def get_image_hexagonalicity(image, rotations=(60, 300)):
+    # init = clock()
```
dneise: such comments should not end up in production code. It is possible to benchmark fuctions in pytests

digicampipe/image/utils.py
```diff
@@ -0,0 +1,416 @@
...
+
+    def Gauss2D(x, y):
+        """ Returns the values of the defined 2d gaussian at x,y """
+        xr = x * np.cos(rot) - y * np.sin(rot)  # X position in rotated coordinates
+        yr = x * np.sin(rot) + y * np.cos(rot)
+        return amplitude * np.exp(-(((xr - Xc) / xsigma) ** 2 +((yr - Yc) / ysigma) ** 2) / 2) + bkg
+    return Gauss2D
+
+
+def FitGauss2D(Data, ip=None):  # from https://github.com/indiajoe/HandyTools4Astronomers
```
dneise: `Data`: should not be capitalized
`ip` is not a very good parameter name, `initial_parameters` should be fine, I guess.

digicampipe/image/utils.py
```diff
@@ -0,0 +1,416 @@
...
+        crop_xmin, crop_xmax = crop_xmax, crop_xmin
+    if crop_ymin > crop_ymax:
+        crop_ymin, crop_ymax = crop_ymax, crop_ymin
+    crop_pixel1 = crop_xmin, crop_ymin
+    crop_pixel2 = crop_xmax, crop_ymax
+    return image[crop_ymin:crop_ymax, crop_xmin:crop_xmax], crop_pixel1, crop_pixel2
+
+
+def average_images(images):
+    if type(images) is not list:
```
dneise: this type check is not needed, since there is an iteration over `images` in line 67.

digicampipe/image/utils.py
```diff
@@ -0,0 +1,416 @@
...
+        print('ERROR in average_images(): images must be a list of filenames or numpy.ndarray')
+        return None
+    nimage = len(images)
+    if nimage == 0:
+        print('ERROR in average_images(): empty list of images')
+        return None
+    if type(images[0]) is not np.ndarray:
```
dneise: this checks, that the type of the first element is correct, but not the other elements. So it is pretty useless... 
Also later there is `images[0].shape` which will fail if `images[0]` does not have a property called `shape`...

digicampipe/image/utils.py
```diff
@@ -0,0 +1,416 @@
...
+        return None
+    first_image_shape = images[0].shape
+    image_average = np.zeros(first_image_shape)
+    for image in images:
+        if type(image) is not np.ndarray:
+            print('WARNING in average_images(): image is not a numpy.ndarray')
+            continue
+        if first_image_shape != image.shape:
+            print('ERROR in average_images(): images are not of same sizes')
+            return None
```
dneise: instead of expecting a `list` of images, this function could expect a 3-dimensional array of images, with shape: (n_images, image_width, image_height). The iteration still works, but it is **ensured** by design that each image in this "stack" of images has the correct type and shape.

I would personally still not do a type-check, but simply document that this function expects a 3d-array. But if one would like to check the type, this should be enough:

    assert images.ndim == 3

dneise: Ah now that I look at it again, one could do this:
~~~python
def average_images(images):
    images = np.asarray(images)  # convert to array, if it not already is an ndarray.
    assert images.ndim == 3        # conversion to 3D array only works, if all images are of equal size.
    return images.mean(axis=-1)
~~~

digicampipe/image/utils.py
```diff
@@ -0,0 +1,416 @@
...
+from skimage.draw import polygon
+import matplotlib.pyplot as plt
+from time import clock
+import cv2
+from scipy import signal, ndimage, optimize
+import numbers
+from photutils import DAOStarFinder
+
+
+def crop_image(image, crop_pixel1, crop_pixel2):
```
dneise: I would propose to make this a 2-parameter function with this interface `image, rectangle`.

digicampipe/image/utils.py
```diff
@@ -0,0 +1,416 @@
...
+        raise AttributeError("invalid crop_pixel position:", crop_pixel1, crop_pixel2, "image size is: ", image.shape)
+    if not (0 < crop_xmax < npixel_x and 0 < crop_ymax < npixel_y):
+        raise AttributeError("invalid crop_pixel position:", crop_pixel1, crop_pixel2, "image size is: ", image.shape)
+    if crop_xmin > crop_xmax:
+        crop_xmin, crop_xmax = crop_xmax, crop_xmin
+    if crop_ymin > crop_ymax:
+        crop_ymin, crop_ymax = crop_ymax, crop_ymin
+    crop_pixel1 = crop_xmin, crop_ymin
+    crop_pixel2 = crop_xmax, crop_ymax
+    return image[crop_ymin:crop_ymax, crop_xmin:crop_xmax], crop_pixel1, crop_pixel2
```
dneise: why does this function also return the rectangle again? I mean sure, it considers boundary effects, but is this actually needed anywhere? 

As a more general thought. Functions can return tuples in python, but often people cannot remember what the order of the stuff inside the tuple is. So often functions returning tuples are hard to use, and if they are used somewhere in code, it is hard to understand what they return, if the variable names are not chosen very good.

digicampipe/image/utils.py
```diff
@@ -0,0 +1,416 @@
...
+
+def crop_image(image, crop_pixel1, crop_pixel2):
+    """
+    :param image: 2D array of the image data with a shape = (vertical size, horizontal size)
+    :param crop_pixel1: iterable of 2 ints: position (pos_x , pos_y) of one corner of the returned region
+    :param crop_pixel2: iterable of 2 ints: position (pos_x , pos_y) of the opposite corner of the returned region
+    :return: the cropped image, the 1st corner and the opposite corner positions
+    """
+    if type(image) is str:
+        image = fits.open(image)[0].data
```
dneise: writing functions, doing one thing only is often a good idea. I reduces maintenance work. This function crops an image, but if it is given a path it also opens that file and retrieves the image from the first HDU, making two assumptions: (a) the file is a fits file, and (b) the image in question is stored as the 1st HDU. 

This function would become easier, if it just accepted an image (i.e. 1 2d-array).

digicampipe/image/utils.py
```diff
@@ -0,0 +1,416 @@
...
+    if type(image) is not np.ndarray:
+        raise AttributeError("image must be a filename or a numpy array.")
```
dneise: I would not do this type check here, later this function does stuff with image, which is only possible if it is a 2d-ndarray. that is the time to raise the exception.... not now.

digicampipe/image/utils.py
```diff
@@ -0,0 +1,416 @@
...
+    if crop_pixel1 is not list and len(crop_pixel1) is not 2:
+        raise AttributeError("crop_pixel1 must be a list of 2 ints", "(position in px of one corner of the crop area).")
+    if crop_pixel2 is not list and len(crop_pixel1) is not 2:
```
dneise: these type checks are not needed, crop_pixel could also be  tuple not a list. Also I think using a `Rectangle` type here, which nicely bundles the 4 numbers together is much easier to handle here, I think.

digicampipe/image/utils.py
```diff
@@ -0,0 +1,416 @@
...
+        raise AttributeError("crop_pixel2 must be a list of 2 ints",
+                             "(position in px of the other corner of the crop area).")
+    crop_xmin, crop_ymin = crop_pixel1
+    crop_xmax, crop_ymax = crop_pixel2
+    if not isinstance(crop_xmin, numbers.Integral) or not isinstance(crop_ymin, numbers.Integral):
+        print('Warning in crop_image(): crop_pixel1 does not contain ints')
+        crop_xmin = int(crop_xmin)
+        crop_ymin = int(crop_ymin)
```
dneise: Often it is better to fail if a parameter has a wrong type, then to do your best to recover the mistake made by the caller. I would just state a "CropRegion" or "CropRectangle" must have integer fields: "top, bottom, left, right" or "x1, x2, y1, y2" or  whatever. If not, we do not convert to integers but simply fail. We expect the caller of this function to provide meaningful inputs and not something, which is almost good and we have to guess what the caller means. 

By simply omitting these type-checks, later this part:

    image[rectangle.bottom:rectangle.top, rectangle.left:rectangle.right]

will throw an exception if e.g. `rectangle.top` is not an integer but a float, or a string or whatever.

digicampipe/fit_cones_position.py
```diff
@@ -0,0 +1,44 @@
...
+    cones_img.plot_cones(output_dir=output_dir, radius_mask=2.1)
+    cones_img.plot_cones_presence(output_dir=output_dir)
+    cones_img.fit_camera_geometry()
+    cones_img.refine_camera_geometry()
+    cones_img.plot_camera_geometry(output_dir=output_dir)
+    return cones_img.pixels_pos_predict, cones_img.pixels_pos_true
+
+
+def get_cones_position(filename, output_dir=None):
+    cones_img = ConesImage(filename, './tests/resources/cones_1509411741-cone.fits')
```
dneise: I think for this path here: `./tests/resources/....`  `pkg_resources` should be used.

digicampipe/image/cones_image.py
```diff
@@ -0,0 +1,705 @@
...
+import numpy as np
+from decimal import *
+import os
+
+
+class ConesImage(object):
+    def __init__(self, image, image_cone=None, output_dir=None,
+                 digicam_config_file='./tests/resources/camera_config.cfg'):
+        """
+        constructor of a ConesImage object.
```
dneise: saying that `__init__` is the c-tor of the object is not necessary.

digicampipe/image/cones_image.py
```diff
@@ -0,0 +1,705 @@
...
+        :param image: fit filename or numpy array containing the lid CCD image.
+        If set to 'test', a test image is created.
```
dneise: image plays a triple role: array, path-string pointing to a fits file, or the special string `test`. This is quite difficult to understand and maintain.

digicampipe/image/cones_image.py
```diff
@@ -0,0 +1,705 @@
+from digicampipe.utils import geometry
+from digicampipe.image.kernels import *
```
dneise: `import *` is not recommended.

digicampipe/image/sky_image.py
```diff
@@ -0,0 +1,555 @@
...
+            raise AttributeError('crop_pixels2 must be a list of pixels.')
+        if len(crop_pixels1) != len(crop_pixels2):
+            raise AttributeError('crop_pixels1 and crop_pixels2 lists must have the same length.')
+        image = fits.open(self.filename)[0].data
+        self.image_shape = image.shape
+        self.crop_pixels1 = []
+        self.crop_pixels2 = []
+        self.sky_images = []
+        self.sky_images_shape = []
+        print('divide', filename, 'in', len(crop_pixels1), 'sub-areas')
```
dneise: this should maybe become a log-statement

digicampipe/image/sky_image.py
```diff
@@ -0,0 +1,555 @@
...
+                try:
+                    header_wcs = fits.open(wcs_file)[0].header
+                    self.wcs = WCS(header_wcs)
+                    # reference point
+                    self.reference_pixel = (header_wcs['CRPIX1'], header_wcs['CRPIX2'])
+                    self.reference_ra_dec = (header_wcs['CRVAL1'], header_wcs['CRVAL2'])
+                    # transformation matrix
+                    self.cd_matrix = np.array(((header_wcs['CD1_1'], header_wcs['CD1_2']),
+                                               (header_wcs['CD2_1'], header_wcs['CD2_2'])))
+                except FileNotFoundError:
```
dneise: Why is it in this situation okay to pass, if the file was not found?

digicampipe/image/sky_image.py
```diff
@@ -0,0 +1,555 @@
...
+                    pass
+            # stars position in image
+            self.stars_pixel = None
+            if success:
+                try:
+                    stars_data = fits.open(corr_file)[1].data
+                    self.stars_pixel = np.array((stars_data['index_x'], stars_data['index_y']))
+                except FileNotFoundError:
+                    pass
```
dneise: Same question as above. Why is it in this situation okay to let this error pass silently?

dneise:
I'd like to fix a few of the issues I had with this PR myself. It's easy to fix them, while I read the code anyway. But the question is if you @yrenier would be fine with me pushing into your branch?

dneise:
Or I just make my modifications in a copy of this branch, and open little tiny PRs, which deal with some of the issues I had. This way @yrenier can nicely decidy if he likes the modification I made or not.

digicampipe/tests/test_lid_ccd.py
yrenier: fixed in commit lid_ccd 2dff226

digicampipe/tests/test_lid_ccd.py
yrenier: That would totally make sense

digicampipe/image/sky_image.py
yrenier: Several images are needed to detect stars (looking for movement).
The only specific work done by LidCCDObservation is to calculate the average of all images, that average being then used to find stars.

digicampipe/image/utils.py
yrenier: I totally agree

digicampipe/image/utils.py
yrenier: The rectangle is returned as it can differ from the given one (f.e. if the rectangle is partially out of the image).
It is then use to translate coordinates from the cropped region to the full image.
I do not like much returning tuples either, but creating a class for that one seemed a bit overkill.

digicampipe/image/cones_image.py
yrenier: Would it be better to create a ConeImageSim class inheriting from ConeImage to handle the "test" case instead ?

digicampipe/image/sky_image.py
yrenier: That file contains stars expected in that region of the sky.
Only created if the coordinates were found.

yrenier:
You can push to my branch or create an other as you wish.
But I'm not sure if I should implement some of the suggestions now or wait you are done to avoid too many interferences.

dneise:
Never wait ... if I want to open a PR into your branch, it is my duty to keep in sync with you and keep the diff between my mods and your code as small as possible, to make review as quick and easy as possible for you. 
If I don't manage that, please just tell me: "the diff is too large, I cannot review it"

pointing_from_stars.py
yrenier: fixed in lid_ccd 28f7e0c

digicampipe/image/cones_image.py
dneise: No, I don't think so. I'll make a proposal in a new PR .. gimme a second.

digicampipe/image/sky_image.py
yrenier: The wcs file is only created if the coordinates could be determined from the list of stars.
If the file is missing, it only means that we got no coordinates.

dneise:
There are some places where functions or class-ctors accept either an image (i.e. a 2d-array) or a path to a fits file. I think this breaks the single responsibility principle, cf. https://en.wikipedia.org/wiki/SOLID_(object-oriented_design)

I'l open a new PR with a proposal to change that for `ConesImage`. After that I will make a purely technical PR, which works on pep8 issues... since my editor is complaining a lot about them, and it distracts me quite a lot.

dneise:
Ah now that I'm thinking about it, I'll start with the pep8 thing

digicampipe/tests/test_lid_ccd.py
yrenier: fixed in 4724d7158293

digicampipe/image/utils.py
yrenier: fixed in 4724d7158293

digicampipe/image/utils.py
yrenier: I did it anyway in 4724d7158293

digicampipe/image/utils.py
yrenier: fixed in 4724d7158293

digicampipe/image/utils.py
yrenier: fixed in 4724d7158293

dneise:
Okay I think we have a problem here with the astrometry.net stuff. When stuff is failing on travis, I cannot fix it locally on my laptop, since I cannot try it since I have no API key. Also it tries to execute the local installation, so this makes it hard to do local testing.

In addition, I have currently problems running the local installation, this will never be found by travis, since travis is only testing the web api. 

So we somehow made testing more difficult not easier, buy supporting both a local installation and the web api

digicampipe/image/sky_image.py
```diff
@@ -0,0 +1,543 @@
...
+                                       scale_high_images_deg=scale_high_images_deg,
+                                       guess_ra_dec = guess_ra_dec,
+                                       guess_radius = guess_radius)
+            if image_shape is None:
+                image_shape = lidccd_image.image_shape
+            elif image_shape != lidccd_image.image_shape:
+                raise AttributeError('all images must have the same size')
+            self.lidccd_images.append(lidccd_image)
+        static_images = []
+        for crop_area_index, crop_area_shape in enumerate(lidccd_image.sky_images_shape):
```
dneise: Here `liccd_image` is just the last image in the list of images, is this intended? It looks like a bug to me

dneise:
I do not understand what `static_images` are are they images containing static, as in white noise?

dneise:
This branch fails locally on my machine:

The failing test is this one:

    tests/test_lid_ccd.py::test_find_stars FAILED

stdout looks like this:
```
------------------------------------------------------------------------------------------------------------ Captured stdout call -------------------------------------------------------------------------------------------------------------
divide /home/dneise/ctasoft/digicampipe/digicampipe/tests/resources/stars_on_lid/stars_1509413134.fits in 2 sub-areas
divide /home/dneise/ctasoft/digicampipe/digicampipe/tests/resources/stars_on_lid/stars_1509411641.fits in 2 sub-areas
divide /home/dneise/ctasoft/digicampipe/digicampipe/tests/resources/stars_on_lid/stars_1509411441.fits in 2 sub-areas
divide /home/dneise/ctasoft/digicampipe/digicampipe/tests/resources/stars_on_lid/stars_1509411010.fits in 2 sub-areas
divide /home/dneise/ctasoft/digicampipe/digicampipe/tests/resources/stars_on_lid/stars_1509411541.fits in 2 sub-areas
divide /home/dneise/ctasoft/digicampipe/digicampipe/tests/resources/stars_on_lid/stars_1509414126.fits in 2 sub-areas
divide /home/dneise/ctasoft/digicampipe/digicampipe/tests/resources/stars_on_lid/stars_1509411110.fits in 2 sub-areas
divide /home/dneise/ctasoft/digicampipe/digicampipe/tests/resources/stars_on_lid/stars_1509410910.fits in 2 sub-areas
divide /home/dneise/ctasoft/digicampipe/digicampipe/tests/resources/stars_on_lid/stars_1509410810.fits in 2 sub-areas
divide /home/dneise/ctasoft/digicampipe/digicampipe/tests/resources/stars_on_lid/stars_1509413730.fits in 2 sub-areas
Reading input file 1 of 1: "/tmp/tmpbjqj7wct/stars.fits"...
Reading input file 1 of 1: "/tmp/tmp35_q8ndx/stars.fits"...
Reading input file 1 of 1: "/tmp/tmp_hq3024t/stars.fits"...
Reading input file 1 of 1: "/tmp/tmpip3va3vg/stars.fits"...
Reading input file 1 of 1: "/tmp/tmpl2cizb9_/stars.fits"...
Reading input file 1 of 1: "/tmp/tmpxs2qgs3h/stars.fits"...
Reading input file 1 of 1: "/tmp/tmpbkcafq3k/stars.fits"...
Reading input file 1 of 1: "/tmp/tmp37utaxqj/stars.fits"...
Reading input file 1 of 1: "/tmp/tmpbpoo9edm/stars.fits"...
Reading input file 1 of 1: "/tmp/tmp82m5jdso/stars.fits"...
Reading input file 1 of 1: "/tmp/tmpc7fibulf/stars.fits"...
Reading input file 1 of 1: "/tmp/tmpyj2cxx6x/stars.fits"...
Reading input file 1 of 1: "/tmp/tmparrmk92w/stars.fits"...
Reading input file 1 of 1: "/tmp/tmpiv7pugg0/stars.fits"...
Reading input file 1 of 1: "/tmp/tmpjfklloz3/stars.fits"...
Reading input file 1 of 1: "/tmp/tmpxy4i4797/stars.fits"...
Reading input file 1 of 1: "/tmp/tmpxenyhqie/stars.fits"...
Reading input file 1 of 1: "/tmp/tmpuy0661iz/stars.fits"...
Reading input file 1 of 1: "/tmp/tmpd_cyodx7/stars.fits"...
Reading input file 1 of 1: "/tmp/tmp6hthf1kl/stars.fits"...
```

stderr looks like this:
```
Traceback (most recent call last):
  File "/usr/lib/python2.7/runpy.py", line 174, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/usr/lib/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "/usr/lib/python2.7/dist-packages/astrometry/util/image2pnm.py", line 323, in <module>
    sys.exit(main())
  File "/usr/lib/python2.7/dist-packages/astrometry/util/image2pnm.py", line 320, in main
    mydir, fix_sdss=options.fix_sdss)
  File "/usr/lib/python2.7/dist-packages/astrometry/util/image2pnm.py", line 248, in convert_image
    (imgtype, errstr) = image2pnm(infile, outfile, sanitized, force_ppm, no_fits2fits, extension, mydir, fix_sdss)
  File "/usr/lib/python2.7/dist-packages/astrometry/util/image2pnm.py", line 174, in image2pnm
    from fits2fits import fits2fits as fits2fits
  File "/usr/lib/python2.7/dist-packages/astrometry/util/fits2fits.py", line 33, in <module>
    from astropy.io import fits as pyfits
  File "/usr/lib/python2.7/dist-packages/astropy/__init__.py", line 112, in <module>
    _check_numpy()
  File "/usr/lib/python2.7/dist-packages/astropy/__init__.py", line 100, in _check_numpy
    from .utils import minversion
  File "/usr/lib/python2.7/dist-packages/astropy/utils/__init__.py", line 18, in <module>
    from .misc import *
  File "/usr/lib/python2.7/dist-packages/astropy/utils/misc.py", line 28, in <module>
    from ..utils.compat.odict import OrderedDict
  File "/usr/lib/python2.7/dist-packages/astropy/utils/compat/__init__.py", line 14, in <module>
    from .numpycompat import *
  File "/usr/lib/python2.7/dist-packages/astropy/utils/compat/numpycompat.py", line 18, in <module>
    NUMPY_LT_1_6_1 = not minversion('numpy', '1.6.1')
  File "/usr/lib/python2.7/dist-packages/astropy/utils/introspection.py", line 154, in minversion
    from pkg_resources import parse_version
  File "/usr/lib/python2.7/dist-packages/pkg_resources/__init__.py", line 25, in <module>
    import zipfile
  File "/usr/lib/python2.7/zipfile.py", line 501, in <module>
    class ZipExtFile(io.BufferedIOBase):
AttributeError: 'module' object has no attribute 'BufferedIOBase'
augment-xylist.c:588:backtick Failed to run command: /usr/bin/python -m astrometry.util.image2pnm --sanitized-fits-outfile /tmp/tmp.sanitized.N19L4H --fix-sdss --infile /tmp/tmpbjqj7wct/stars.fits --uncompressed-outfile /tmp/tmp.uncompressed.wju5xl --outfile /tmp/tmp.ppm.4AXo1Y --ppm
 ioutils.c:567:run_command_get_outputs Command failed: return value 1
```

and this error is repeated as many times as we have test images.

This looks like I am looking at a error with astropy.net and not at an error in our code. 

**But**
The code worked before I made my pep8 stuff, and since I did not touch the astrometry stuff on my machine today, it cannot really be a bug in astrometry, but it must be in our stuff.

Does it work on your machine?

dneise:
Since the traceback starts with `File "/usr/lib/python2.7/runpy.py", line 174, in _run_module_as_main` I am pretty sure, this traceback comes not from our code, since our code is executed by my anaconda python 3 and not by my system python 2.7 ... 

However, when I call my local `solve-field` with wrong input: like this:
```
(digicampipe) dneise@lair:~/ctasoft/digicampipe/digicampipe$  solve-field fooo 
Reading input file 1 of 1: "fooo"...
ERROR: Image type not recognized: Unknown image type "["cannot open `fooo' (No such file or directory)"]"
augment-xylist.c:588:backtick Failed to run command: /usr/bin/python -m astrometry.util.image2pnm --sanitized-fits-outfile /tmp/tmp.sanitized.tj7oMp --fix-sdss --infile fooo --uncompressed-outfile /tmp/tmp.uncompressed.zsAZKd --outfile /tmp/tmp.ppm.6LhAJ1 --ppm
 ioutils.c:567:run_command_get_outputs Command failed: return value 255
```

I get an error that looks very similar to the errors I've shown above. So I am ... pretty sure, that my local installation of astrometry.net might have a problem or so ...

dneise:
I have problems understanding the code in many places. Mostly this is due to too long functions. I will start now breaking some functions apart.

yrenier:
The astrometry stuff is working locally for me.
I have seen it uses python 2 to do some stuff (f.e. extract sources), maybe some pakages need to be installed system wide for astrometry to work locally.
About the tests made more complicated with both interfaces to local and remote astrometry, I as thinking to only test for remote, and let the user deal with the local installation if he wants it.

yrenier:
static_image is just an average of all the images.
I remove that static image from individual images and what remains are considered as stars (as stars are moving from one image to the next).

yrenier:
otherwise getting an API key for astrometry is very simple:
1. login at http://supernova.astrometry.net/signin/
2. you can see your API key at http://supernova.astrometry.net/api_help (in green)

digicampipe/image/sky_image.py
yrenier: I'm not so sure how to do that :-/

yrenier:
I believe I fixed all requested changes except about logging instead of print XD

dneise:
I believe there are still some issues hidden, which I only realize step by step. I'll write them up, as soon as I am in the office

dneise:
First, I have a question again regarding `calculate_galactic_coordinates_local`.
There are 3 try-except blocks. Let's assume one of the files is missing say the ".wcs" file. Is the result useful then? 

Or is it like this:
 * Either all files are there, then the result is useful, or
 * at least one of the files is missing, then we cannot use the rest of the information anyway?

dneise:
Here are some of the issues, I still stumble over:

* Tests are too slow -> mock `solve-field` and reduce cone_simu in size?
* More tests showing what all the new functions and classes are supposed to do, would make it easier to understand them (apart from the important case of showing that they work)
* functions are too long, hard to understand and reason about. (Also sometimes hard to test)
* Functions have side effects (makes hard to test)
    Analysis functions which also plot or also store the results, have side effects. (I can give examples)
* Functions with boolean parameters, doing two different things. Sometimes hard to understand. 
* unused functions: `crop_image` is not used anywhere? (did I do that by incorrect merging?)
* A Hexagon can be specified by: "center, radius, angle": 4 parameters, but e.g. in `set_hexagon` it is represented by 6 paramters. Why? Are maybe also "skewed" hexagons allowed?
* There are still many commented out code-lines, looks like not yet ready. 
* There is a "Not implemented yet" still in the code.

dneise:
In `calculate_galactic_coordinates_local` is this code:

```python
                self.wcs = WCS(header_wcs)
                # reference point
                self.reference_pixel = (
                    header_wcs['CRPIX1'], header_wcs['CRPIX2'])
                self.reference_ra_dec = (
                    header_wcs['CRVAL1'], header_wcs['CRVAL2'])
                # transformation matrix
                self.cd_matrix = np.array((
                    (header_wcs['CD1_1'], header_wcs['CD1_2']),
                    (header_wcs['CD2_1'], header_wcs['CD2_2'])
                ))
```

After this code is executed, this `assert (self.cd_matrix == self.wcs.wcs.cd).all()` can be done to show that `self.cd_matrix` is already contained inside `self.wcs`. So this is redundant.

yrenier:
About `calculate_galactic_coordinates_local`:
The most important file is the wcs as it contains the fitted coordinates. 
The axy file is containing what was found as stars (this stars are then compared to catalog to determine coordinates)
The xyls file is containing the stars from the catalog around the found coordinates.
Right now, if there is no wcs file, the result is not useful. However once coordinates are found for at least one region in the image, I'm planning to implement later an additional global fit using the stars from other regions too.

yrenier:
About mocking solve-field, it is probably the way to go if we implement tests for the local install.
However for the web-service, the test are quite fast already as the solving is already done and we only dowload the results.

yrenier:
about set_hexagon() : indeed the hexagons are allowed to be skewed.
Defining them using 2 consecutive radius is probably not the most straightforward way, but it was the easiest way for that analysis (as radius can be easily determined from the 2D FFT)

yrenier:
The code for the cones position determination is done, but the analysis on coordinates determination from stars can still be improved and so it will probably continue to evolve. As long as the code is working (even sub-optimally) I would not have thought that would be a problem.

yrenier:
For the "functions with boolean parameters, doing two different things. Sometimes hard to understand."
I tried to fix that yesterday for the plotting functions of ConesImage. Or maybe you refer to something elese ?

dneise:
> About mocking solve-field, it is probably the way to go if we implement tests for the local install.
However for the web-service, the test are quite fast already as the solving is already done and we only dowload the results.

Ah I forgot one important point about this matter, sorry. So, tests are executed by fellow developers and users alike. For users its fine if tests take some time, the execute them once after download to be sure everything works on their machine. ok. 
Fellow developers execute tests *all the time*, for them it is important that they can execute the tests all the time, to see they are still passing.

Now imagine a digicampipe developer who has nothing to do with lid ccd image analysis (me). They have neither installed astrometry.net nor do they have a supernova API key. And even if they have one, you explained nicely, why the tests will still take so long for them, they essentially are useless. Can they execute the tests? No.

If they install astrometry.net locally, and something does not work (happens to me yesterday, and is still the case), does it help them that travis is testing the web-api? No. So they cannot develop, until they have setup astrometry.net correctly, which is not actually needed for their contribution to digicampipe.

I think this is suboptimal.

dneise:
I tried to speed up the cones_simu by specifying a smaller camera. With only 20 pixels the simulation and the test, I figured, should run faster. But I ran into problems with the camera configuration.

What I did was this:
 * take the camera config file and cut away so many lines, that I end up with only 20 pixels.
 * modify the test like this:

```python

from pkg_resources import resource_filename
small_camera_config_file = resource_filename(
    'digicampipe',
    'tests/resources/camera_config_20_pixel.cfg')


def test_cone_simu():
    from digicampipe.image.cones_image import ConesImage, cones_simu
    from digicampipe.image.cones_image import simu_match
    from digicampipe.image.cones_image import get_pixel_nvs

    # create an image  with a geometry compatible to the camera with
    # known angle, spacing between pixels etc...
    test_image, true_positions = cones_simu(
        pixels_nvs=get_pixel_nvs(small_camera_config_file),
        offset=(-4.3, -2.1),
        angle_deg=10,
        pixel_radius=35,
    )
    cones_img = ConesImage(
        test_image,
        digicam_config_file=small_camera_config_file)
    cones_img.get_cone(radius_mask=2.1)
    cones_img.fit_camera_geometry()
    cones_img.refine_camera_geometry()

    assert simu_match(cones_img, true_positions, std_error_max_px=0.5)
```

But I got an error from `cts_core.camera.Camera`, so apparently one cannot simply make a different camera, by cutting out lines from the config file ...

dneise:
> The code for the cones position determination is done, but the analysis on coordinates determination from stars can still be improved and so it will probably continue to evolve. As long as the code is working (even sub-optimally) I would not have thought that would be a problem.

When I said, "the tests are slow" I was not referring to the performance of the analysis, but more on the scope of the tests. A test can act on a small example, to show that something works. In case the analysis scales linearly, reducing the data-sample a test acts on by a factor of ~100 can make a large difference.

See here for example the durations of the slowest 5 tests:
```
197.79s call     digicampipe/tests/test_cone_simu.py::test_cone_simu
51.58s call     digicampipe/tests/test_lid_ccd.py::test_find_stars
2.28s call     digicampipe/tests/test_zfits_event_source.py::test_and_benchmark_simple_event_source
1.28s call     digicampipe/tests/test_zfits_event_source.py::test_and_benchmark_expert_event_source
0.02s call     digicampipe/tests/test_read_raw_events.py::test_adc_samples
```

I thought, reducing the size of  the cone simulation from ~1000 pixels to ~10 pixels might reduce runtime down to ~2sec which would be very nice for me.

But it seems to be not easy to reduce the size of the data sample

dneise:
If we cannot speed up the tests, or if we say: 

We could speed up the tests, but we don not really know how, and it takes a lot of our time, and we do not want that right now.

Then I would propose to either mark these tests as "slow" or as "astrometry" or so, as explained here: https://docs.pytest.org/en/latest/example/markers.html

People who do not have astrometry.net locally installed or for other reasons do not want to execute these tests all the time. Can then simply deselect them like

    pytest -m "not astrometry"

Would you be open for that?

yrenier:
nicely spotted, I fixed the redundancy in SkyImage.

yrenier:
You made your point very clear about speed of test, I an only agree.
I tried also to make the test of the cones with much less pixels, but I did not have more luck with than you did.
Giving the user the possibility to the user to pass the slow test makes sense, than for implementing it !

yrenier:
I was thinking of creating a dummy google account so we could use it for all astrometry stuff making the API key public. That would solve most of the problem you mentioned.

yrenier:
Thank you very much for your patience, I really don't feel like it was a waste of my time as I believe the code is in a much better state as it was before and I learned a few thing in the process :-D

dneise:
>  I really don't feel like it was a waste of my time as I believe the code is now in a much better state and I learned a few thing in the process :-D

Thanks for these nice words. I also learned quite some stuff, especially about some astronomy packages.

PR #68, opened 2018-01-03
Lid ccd imports

dneise:
replace asterisk imports by explicit imports.

Also sort imports according to:
 * std-libary
 * dependencies, we have not written ourselves
 * our own stuff

### DIFF
--- a/digicampipe/image/cones_image.py
+++ b/digicampipe/image/cones_image.py
@@ -1,6 +1,6 @@
 import os
 import decimal
-from decimal import Decimal
+from decimal import Decimal, ROUND_HALF_EVEN
 import numpy as np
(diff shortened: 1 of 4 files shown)


dneise:
I searched for the error on the web and found something about `import cv2` which needs to be above or below some other import statement:
https://github.com/tensorflow/models/issues/523

dneise:
yes finally ... something seems to be broken somewhere in opencv or so ... so we need to take care in which order it is imported ... what a pity.

dneise:
Ah in case you want to merge this PR, please keep in mind that the merge might create conflicts in PR #69. No problem, I just have to resolve them....

PR #69, opened 2018-01-03
Lid ccd cones image test

dneise:
This is a proposal, how to remove the special 'test' string parameter from the ConesImage c-tor. 

I think the ConesImage c-tor should accept only an image as a parameter to its contructor. If we want to test it, we give it a test image. Of course we have to *make* the test_image before we can give it. So it looks like it is a bit more work, but I think code under test should not "know" it is under test, so it should not have dedicated `if`s which are needed for testing.

It still has a special parameter, which is the position of the `true_pixel_positions`, but I see already how to remove that ...

### DIFF
--- a/digicampipe/fit_cones_position.py
+++ b/digicampipe/fit_cones_position.py
@@ -1,8 +1,18 @@
-from digicampipe.image.cones_image import ConesImage
+from digicampipe.image.cones_image import ConesImage, cones_simu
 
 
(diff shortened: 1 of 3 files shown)


dneise:
Okay so this works ... please have a look how you like it.

dneise:
Ah in case you want to merge this PR, please keep in mind that the merge might create conflicts in PR #68. No problem, I just have to resolve them....

Issue #73, opened 2018-01-16
cannot execute `pipeline_crab` on crab events


dneise:
Today I finally found time to download the crab files using Etienne's script.
The files are named: 
```
SST1M01_20171030.000.fits.fz
....
SST1M01_20171030.091.fits.fz
DigicamSlowControl_20171030_004.fits
DigicamSlowControl_20171030_007.fits
DigicamSlowControl_20171030_009.fits
DriveSystem_20171030_001.fits
DriveSystem_20171030_003.fits
SafetyPLC_20171030_000.fits
SafetyPLC_20171030_005.fits
```

I call the script like this:
```
(digicampipe) dneise@lair:~/ctasoft/digicampipe$ python pipeline_crab.py \
    -p /home/dneise/ctasoft/sst1m_crab/ \
    -o /home/dneise/ctasoft/sst1m_crab/out/
```

And get this error:
```
Traceback (most recent call last):
  File "pipeline_crab.py", line 36, in <module>
    dark_baseline = np.load(directory + 'dark.npz')
  File "/home/dneise/anaconda3/envs/digicampipe/lib/python3.5/site-packages/numpy/lib/npyio.py", line 370, in load
    fid = open(file, "rb")
FileNotFoundError: [Errno 2] No such file or directory: '/home/dneise/ctasoft/sst1m_crab/dark.npz'
```

dneise:
So obviously I am missing a file called "dark.npz". Should I have downloaded this file? 
Can I generate this file somehow?

And is there already some docu about this, which I simply should read? :-D

calispac:
this file is supposed to be generated with : https://github.com/calispac/digicampipe/blob/master/dark_evaluation.py

It basically reads the data taken with the lid closed compute the baseline and creates this .npz to store it. This baseline (in dark condition) is used to compute the baseline shift, used for : 
- nsb computation
- gain drop computation

Here I see two issues at the moment : 

1.  There is no way to know if an event is taken with lid closed or opened (at least to my knowledge) from the .zfits or the slow control data. (only from the logbook)
2. As you experienced it is not documented how to generate this file. It was proposed to add this computation step into the crab_pipeline.py

Etienne12345:
One can retrieve the status of the lid from the slow control data. I am actually doing it with the metadata that is extracted "per raw data file". Should I send an email to Yves to ask him to add that to the pipeline ? I can explain him where the value should be retrieved from.

dneise:
[...]

PR #77, opened 2018-01-17
Dark evaluation issue

### LINKED ISSUE #75: Cannot execute `dark_evaluation`
As mentioned in #74 I think users of scripts should not have to modify them to get them to work, but to call them with command line arguments.

Nevertheless, I tried to execute `dark_evaluation.py` on the crab files I downloaded using Etiennes script.

This is the diff of `dark_evaluation.py`
```diff
diff --git a/dark_evaluation.py b/dark_evaluation.py
index fe2643a..d242d0e 100644
--- a/dark_evaluation.py
+++ b/dark_evaluation.py
@@ -7,14 +7,24 @@ from digicampipe.utils import geometry
 import matplotlib.pyplot as plt
 import numpy as np
 
+import pkg_resources
+import os
 
 if __name__ == '__main__':
     # Data configuration
 
-    directory = '/home/alispach/data/CRAB_01/'
-    filename = directory + 'CRAB_01_0_000.%03d.fits.fz'
+    directory = '/home/dneise/ctasoft/sst1m_crab/'
+    filename = directory + 'SST1M01_20171030.%03d.fits.fz'
     file_list = [filename % number for number in [2]]
-    camera_config_file = '/home/alispach/ctasoft/CTS/config/camera_config.cfg'
+
+    camera_config_file = pkg_resources.resource_filename(
+        'digicampipe',
+        os.path.join(
+            'tests',
+            'resources',
+            'camera_config.cfg'
+        )
+    )
     dark_filename = 'dark.npz'
 
     trigger_filename = 'bias_curve_dark.npz'
@@ -27,14 +37,13 @@ if __name__ == '__main__':
     digicam_geometry = geometry.generate_geometry_from_camera(camera=digicam)
 
     # Define the event stream
-    data_stream = event_stream(file_list=file_list, expert_mode=True, camera_geometry=digicam_geometry)
+    data_stream = event_stream(file_list=file_list, expert_mode=True, camera_geometry=digicam_geometry, camera=digicam)
     data_stream = filter.set_patches_to_zero(data_stream, unwanted_patch=unwanted_patch)
     # Fill the flags (to be replaced by Digicam)
     data_stream = filter.filter_event_types(data_stream, flags=[8])
 
     data_stream = save_bias_curve(data_stream,
                                   thresholds=thresholds,
-                                  camera=digicam,
                                   blinding=blinding,
                                   output_filename=directory + trigger_filename,
                                   unwanted_cluster=unwanted_cluster)
```

... wait am I probably working with an old version here or so? Let me reinstall everything quickly.

dneise:
This strives to fix #75 

Okay .. I'm kind of happy with this ... please review and see if you also like it.


dark_evaluation.py
```diff
@@ -27,23 +27,22 @@
     digicam_geometry = geometry.generate_geometry_from_camera(camera=digicam)
 
     # Define the event stream
-    data_stream = event_stream(file_list=file_list, expert_mode=True, camera_geometry=digicam_geometry)
+    data_stream = event_stream(file_list=file_list, expert_mode=True, camera_geometry=digicam_geometry, camera=digicam)
+    data_stream = r0.fill_trigger_input_7(data_stream)
```
calispac: this should be done after "setting patches to zero"

dneise: thx

dark_evaluation.py
```diff
@@ -4,6 +4,8 @@
 
     Example call:
     dark_evaluation.py . path/to/SST1M01_20171030.002.fits.fz
+    speed is ~100 events/second.
+    duration ~2minutes.
```
calispac: I am ok with adding duration but it could depend on how many files (more precisely events there is). 
Maybe one should add in parenthesis : '(for k events)'

dneise: I'll remove it .. I've never seen such info in a doc-string ... that was a bad idea ... sorry.

dneise:
Ah ... tests are missing. But we can also do that in a different PR

dark_evaluation.py
```diff
@@ -1,23 +1,43 @@
...
 
-    directory = '/home/alispach/data/CRAB_01/'
-    filename = directory + 'CRAB_01_0_000.%03d.fits.fz'
-    file_list = [filename % number for number in [2]]
-    camera_config_file = '/home/alispach/ctasoft/CTS/config/camera_config.cfg'
-    dark_filename = 'dark.npz'
 
-    trigger_filename = 'bias_curve_dark.npz'
     thresholds = np.arange(0, 400, 10)
     unwanted_patch = [306, 318, 330, 342, 200]
```
calispac: actually this is also supposed to be an option. it was the patches that were not working during the Crab observations.
I suppose we can live it like this since all the data we have is containing this problem.

dark_evaluation.py
```diff
@@ -27,45 +47,38 @@
...
 
 
-    ## Filter the events for display
+    # Filter the events for display
+if __name__ == "__main__":
+    args = docopt(__doc__)
+    print(args)
+    main(
+        output_directory=args['<output_directory>'],
+        files=args['<files>']
```
calispac: can we have default values from one of the test files ?

dneise: erm .. not with docopt ... i.e. it would be very ugly with docopt ... let me show you.

dneise: But where do you want to write the output? the current working directory, so the dot `.`?

dneise: Ah .. there is another "problem". I implemented these arguments as *mandatory* arguments. So it is as if I said:

This program cannot possibly guess any useful default input. It cannot guess where you want the output to be and it cannot guess where you have your SST1M files lying around. 

Like `mv` cannot guess what you want to move where. And `tar` cannot guess what you wanna compress.

Can I ask you, why you want to say to the user: "Hey, in case you have no idea what to do: Just use one of the example files ... it's fine".

calispac: yes current directory seems ok ? ok should we have `null/` folder in the project?... maybe not.
I would think it is good if we use those template files such that we can launch easily "demos".

Also maybe adding python `dark_evaluation.py a_template_file` command to travis could help to check if it at least runs?

dneise: Okay have a look, now these are not mandatory arguments anymore but options, with default values

baseline.py
```diff
@@ -0,0 +1,66 @@
...
+    data_stream = event_stream(
+        file_list=files,
+        expert_mode=True,
+        camera_geometry=digicam_geometry,
+        camera=digicam
+    )
+    data_stream = filter.set_patches_to_zero(
+        data_stream,
+        unwanted_patch=unwanted_patch)
+    data_stream = r0.fill_trigger_input_7(data_stream)
```
calispac: This is not necessary if you only want to compute the ''dark'' baseline. Only necessary for trigger related measurements.

baseline.py
```diff
@@ -0,0 +1,66 @@
...
+    camera_config_file = pkg_resources.resource_filename(
+        'digicampipe',
+        path.join(
+            'tests',
+            'resources',
+            'camera_config.cfg'
+        )
+    )
+
+    thresholds = np.arange(0, 400, 10)
```
calispac: Same here ! variable not used

baseline.py
```diff
@@ -0,0 +1,66 @@
...
+
+    digicam = Camera(_config_file=camera_config_file)
+    digicam_geometry = geometry.generate_geometry_from_camera(camera=digicam)
+    data_stream = event_stream(
+        file_list=files,
+        expert_mode=True,
+        camera_geometry=digicam_geometry,
+        camera=digicam
+    )
+    data_stream = filter.set_patches_to_zero(
```
calispac: this should probably replaced by `set_pixels_to_zero()`

baseline.py
```diff
@@ -0,0 +1,66 @@
...
+        path.join(
+            'tests',
+            'resources',
+            'camera_config.cfg'
+        )
+    )
+
+    thresholds = np.arange(0, 400, 10)
+    unwanted_patch = [306, 318, 330, 342, 200]
+    unwanted_cluster = [200]
```
calispac: make an unwanted_pixel list + introduce it as input arguments then use it in `set_pixels_to_zero()`

calispac: remove unwanted_patch, unwanted_cluster

baseline.py
dneise: nice, so it gets faster .. thx .. done

baseline.py
dneise: well spotted thx... done

baseline.py
dneise: Can you explain why?

baseline.py
dneise: can you explain why?

calispac: basically in dark_evaluation.py script you want to extract all info regarding : dark events 

1. baseline, baseline std
2. trigger rat per patch
3. rate scan

but for baseline.py I assume you want it to just give you this dark.npz for running the crab_pipeline.py ? 
in that case you don't need the trigger information (in the R0Camera Container). Therfore only pixel images are important. The user would mask some failing part of the camera with pixels.
Sorry if it is not clear... gotta go. Will explain better :)

dneise: > but for baseline.py I assume you want it to just give you this dark.npz for running the crab_pipeline.py ?

Yes exactly. That seemed to be useful.

> in that case you don't need the trigger information (in the R0Camera Container). Therfore only pixel images are important. The user would mask some failing part of the camera with pixels.

Aha .. I think I found my mistake. So I assumed (wrongly) that this means:
```
    unwanted_patch = [306, 318, 330, 342, 200]
    unwanted_cluster = [200]
```
these 5 patches and that cluster ... they are somehow broken and we do not want to look at them.
I assumed all of these pixels have a problem, so it is good to remove these pixel from all analysis.

This assumption was apparently wrong.

So I have a question: Why are these patches unwanted, and why is that cluster unwanted?

dneise: Was that just an example? To show that patches and clusters can be removed?

calispac: > Was that just an example? To show that patches and clusters can be removed?

Yes basically cluster 200 or more precisely cluster_7 contains patches [306,318,330,342,200]  (N.B. there is not 7 patches because it is at the camera border.)

Right now `set_*_to_zeros()` is just, as the name suggests, setting zero values to the raw data (sort of getting rid of noisy components). 

We might want to keep in mind that sometimes we could do some kind of flagging on the status of the pixels/patches/clusters. As there could be different actions to take depending on the hardware issue. For instance you could have : pixel 0 noisy and patch 0 (containing pixel 0) so you could decide to do : 

1. 
```python 
noisy_pixel = 0
image[noisy_pixel] = 0
```
2.
```python 
noisy_pixel = 0
image[noisy_pixel.patch.pixels] = 0
```
3.
```python 
noisy_pixel = 0
image[noisy_pixel.patch.pixels] = 0
trigger_in[noisy_pixel.patch] = 0
```
4.
```python 
noisy_pixel = 0
image[noisy_pixel] = 0
trigger_in[noisy_pixel.patch] = 0
```

This depends on what is the hardware issue.

baseline.py
```diff
@@ -0,0 +1,50 @@
+#!/usr/bin/env python
+'''
+Extract baseline from observations.
+
+    Example call:
+    baseline.py ./baseline.npz path/to/SST1M01_20171030.002.fits.fz
+    baseline.py --unwanted_pixels=1,2,3 delete_me.test ../sst1m_crab/SST1M01_20171030.002.fits.fz
```
calispac: what is the delete_me.test for ?

dneise: it is the output file path

dneise:
This fixes #73 when merged

dneise:
Okay .. test are passing ... I think we can merge this

dneise:
There is still work to be done .. but maybe its fine for now...

PR #79, opened 2018-01-17
pipeline_crab.py did not work

dneise:
So just like the other script. Also this script did not work.

I repaired it and gave it command line arguments


dneise:
Have look if you like it. 

Keep in mind. **It did not work before** this PR is not about giving it a different command line interface.

So you could say: Then why did you change the command line interface? 
The answer is: Because before that, the script was looking for files, I do not even have. Only after I modified the CL-interface, I was able to even start debugging.

The amount scripts which are not running, when I try them out is scary. 
Either I am doing something fundamentally wrong and the scripts actually do work, but I use them wrongly. Then this PR is rubbish and must not be merged.

Or the scripts really do not work. In that case I wonder why I am the one running into this? I assumed everybody is constantly using them.

Since this seems not to be the case, we have to think about our testing strategy. 

We do not have to write many tests, but we should write important tests. That means if a script is done ... and then nobody uses it for a while .. this script should still be tested, so that after a month or two it still works and people can actually rely on stuff.

All sphinx documentation and doc-strings do not help, when the software is not running.

calispac:
basically the scripts in the main folder are intended as "example" scripts. They were never tested by more than 2 users and definitely not by a CI system. 
Concerning tests I would prefer that we write tests for the functions found in `digicampipe/digicampipe/` at first rather than for these scripts.

What do you propose for testing for instance `dark_evaluation.py` ? Do you want to do the `assert` thing with pytest or is it ok for you to ask travis to just run it with the default values ?

digicampipe/calib/camera/dl2.py
```diff
@@ -19,25 +19,19 @@ def calibrate_to_dl2(event_stream, reclean=False, shower_distance=80*u.mm):
...
-                                                             distance=shower_distance)
-                    dl1_camera.cleaning_mask = dl1_camera.cleaning_mask & mask_near_center
-                    image[~dl1_camera.cleaning_mask] = 0
-                    moments = hillas.hillas_parameters(geom, image)
-                else:
-                    moments = moments_first
-            except HillasParameterizationError:
-                print('could not recompute Hillas')
-                moments = None
+            moments_first = hillas.hillas_parameters(geom, image)
```
calispac: I don't get why the `try` --> `except` was removed. what if we get a HillasParameterizationError?

dneise: 2 Reasons:
 1. HillasParameterizationError was a SyntaxError. "HillasParameterizationError" was simply unknown. So once the except was reached a SyntaxError was raised. 
 2. When `moments=None` later code is trying to do access `moments.size.shape` or so .. I do not remember. That did not work since moments was None, and None has no attribute `.size` or however it was called.

So I caught the exception deeper down in `hillas.hillas_parameters` ... but obviously I did not push that file ... damn it.

calispac: Ok but we should still have an except here. Since sometimes the `hilllas_parameters()` function can fail in that case we don't want the pipeline to stop. So maybe setting it to None is wrong but we rather set it a 0/nan `MomentParameters` container ?

calispac: Or maybe just jump to the next iteration with `continue` ?

dneise:
One can test top-down or bottom-up or a mixture or just to some random sampling and just test what seems to be important. That are are more or less valid approaches.

Sometimes code is written very tightly coupled which makes it hard to test. Some developers do not have testing in mind so much. That's life. For me as an outsider it is often hard if not impossible to understand what is coupled how, so testing a function (as in unit testing) for which I am failing to even generate some test input is impossible.

So I simply cannot test that code. I have no idea how to test something, that I do not understand. 
In order to understand it ... I have to execute it ... and see what it does ... maybe throw a couple of print statements .. when I am lost ... 
But this all is entirely impossible when these "examples" do not work.

In my understand these are not examples. These are the most important part of the software.
I mean, there is a difference between a library and an application. 

A library is just a bunch of functions or classes that are "somehow useful". They might come with some examples, that show possible users of the library, how to use it. These examples should be tested, since if the user cannot trust that the library works as he expects it to work, he cannot use it.

An application (in our case) is a couple of programs, intended to be used by other people. These programs hopefully make use of good code design, so they have a couple of functions and classes in the background, which help developers to build the final application. But all these functions ... are they important? No ... not unless the application works.

So in both cases, tests have to make sure that the examples or the applications however we want to call them work. If you want to call digicampipe a library or a group of analysis applications. We have to make sure that the stuff works. And we have to make sure that what worked two weeks ago, still works in two weeks. I assume we do not have a bunch of developers in the basement who just wait to repair stuff the got somehow destroyed within the last weeks.

It is in my opinion not helpful to say: "yes, but we are still in development phase, we have to be able to break stuff". 
Sure ... if we think some application is done wrong and we want to do it differently, then we do it. But then we hopefully do not leave an application lying around which is sure **note to work**. So that other people who come along and trying to execute that, have to wonder why it does not work for them.

This needs also some amount of courage, since deleting stuff seems like wasting the time, that was investing in making it in the first place.

One more about testing the end-product. It is very well possible to write a test, which cannot (yet) work since the application is not (yet) done. Such tests are good guidelines for other developers, since they show what is still needed. 
Such tests can be marked as "skipped" (yellow in pytest) and every developer should try to "unskip" tests .. Once everything is green. The product is ready to be shipped.

Such a procedure of writing tests first is called "test driven development", there are many books about this. It leads to an end-product that has all the tests its needs. And that can never break without us knowing it. But this procedure needs time to learn. There are courses for software devs, which cost quite some money, where you can learn TDD. 

Anyway. There is no need to go full scale TDD here. But testing the important parts, so that fellows can trust those parts that were already done some weeks ago, we can certainly do.

Of course this discussion cycles also around the question what is "important" for digicampipe. I cannot judge that. But I would assume these things are important:

 - to extract images from waveforms
 - to apply some kind of cleaning to these images
 - to extract some kind of image parameters from these images, e.g. Hillas
 - to estimate the energy, type and direction of the primary.

And I assume for each of these steps, one can define some (1..10) sub steps, which can also be nicely tested.

digicampipe/calib/camera/dl2.py
dneise: Oh! I know why I did not push the file .. It is in ctapipe.

dneise: Ok now I did not fix it in ctapipe, but in digicampipe.

dneise: I got confused about what is what. Sometimes I see something funny and want to fix it, and then I see it is part of "CTS" .. this is confusing for me.

digicampipe/calib/camera/dl2.py
```diff
@@ -2,6 +2,26 @@
 import numpy as np
 import astropy.units as u
 
+def hillas_parameters(geom, image):
+    unit = geom.pix_x.unit
+    try:
+       moments = hillas.hillas_parameters(geom, image)
+    except hillas.HillasParameterizationError:
+        return hillas.MomentParameters(
+            size=None,
```
calispac: Sorry but why None and not np.nan ?

dneise: I don't know

dneise: I do not know what is more apropriate here.

Etienne12345: I'd stick to None, as it suggests that it was not initialized. np.nan is ambiguous between "not initialized" and "wrongly computed" IMO

dneise:
Okay now this call:

    ./pipeline_crab.py --baseline_path=../sst1m_crab/dark.npz --outfile_path=./hillas_output.txt ../sst1m_crab/SST1M01_20171030.01*

creates and output file of length 491. So only 491 events survive out of these 10 runs. Is that expected?

pipeline_crab.py
```diff
@@ -71,30 +82,35 @@
...
+        time_integration_options['peak'],
+        time_integration_options['mask'],
+        time_integration_options['mask_edges']
+    ) = utils.generate_timing_mask(
+        time_integration_options['window_start'],
+        time_integration_options['window_width'],
+        peak_position
+    )
 
     # Image cleaning configuration
```
calispac: as these are image cleaning parameters maybe it is better to set them with command line

calispac:
> creates and output file of length 491. So only 491 events survive out of these 10 runs. Is that expected?

It all depends on the cuts applied to image cleaning and the trigger threshold. I remember we had something around ~10Hz "shower" rate.

pipeline_crab.py
dneise: Yep, can be done .. not in this PR. This PR tries to make this script runnable.

digicampipe/calib/camera/dl2.py
dneise: It might be that the size is zero and for this reason all the other parameters cannot be calculated be calculated. But I am guess here, one would have to study the `hillas_parameter` function in detail and find the reason for the exception.

calispac: Yes that's why it fails! 
https://github.com/cta-observatory/ctapipe/blob/b7a11b8c06c8f0140338974c262f58e650e6fcb7/ctapipe/image/hillas.py#L547-L548

calispac: so should we just skip the event instead of returning a nan/None Container ?

dneise: If we think of these generators as boxes in an hypothetical analysis process. And one of these boxes skips events, then all following boxes never see these events ... this includes boxes doing statistics ... boxes which might be an event display ... i don't know what. 

This might be pretty surprising for people.

I mean a box called "filter_this_and_that" ... clearly skips events. but a box called "calibrate _to_dl2" .. I'm not sure.

I guess this is a design question. One might say: All boxes just **add** information to the event. 
Some kind of "sink" in the end writes the results to disk, whatever was defined to be a result.

But we can also say: Boxes do whatever they like with the event_stream and people just have to read the docu, or the implementation of each box in case there is no docu.

dneise: I think however, size should be zero, because this is really the case here.

dneise:
Okay. If this is your decision... can do that as well

PR #81, opened 2018-01-18
utility class "Camera" for "cts_core.camera.Camera"

### LINKED ISSUE #80: Camera - config_file and geometry
In many places I see code like this:
```python
camera_config_file = pkg_resources.resource_filename(
        'digicampipe',
        path.join(
            'tests',
            'resources',
            'camera_config.cfg'
        )
    )
digicam = Camera(_config_file=camera_config_file)
digicam_geometry = geometry.generate_geometry_from_camera(camera=digicam)
```

And later I see functions, which need both `camera` and `camera_geometry` as parameters.

What if we would make a little utility class like this? maybe in `ctapipe.utils`.
```python
import pkg_resources
from os import path

from cts_core import camera
from digicampipe.utils import geometry


class Camera(camera.Camera):
    def __init__(self, *args, **kwargs):
        if not args and kwargs.get('_config_file', None) is None:
            kwargs['_config_file'] = pkg_resources.resource_filename(
                'digicampipe',
                path.join(
                    'tests',
                    'resources',
                    'camera_config.cfg'
                )
            )
        super().__init__(*args, **kwargs)

        self.geometry = geometry.generate_geometry_from_camera(camera=self)
```

This little helper class has two advantages, without touching the code in CTS.
 1. if no `_config_file` is given, it just falls back to the `camera_config.cfg` which is provided with digicampipe.
 2. It has an *additional* attribute `.geometry`.

So we can now slowly start porting functions which at the moment have this interface:
```python
def some_function(camera, camera_geometry):
    # do something with camera and camera_geometry
```

to this interface:
```python
import warnings

def some_function(camera, camera_geometry=None):
     if camera_geometry is None:
         camera_geometry = camera.geometry
    else:
        warnings.warn("camera_geometry is deprecated. Use utils.Camera instead", DeprecationWarning)
    # do something with camera and camera_geometry
    # this code does not even realize that something was changed above it.
```

And then at some point when we assume every part of the code was ported to the new interface:
```python
def some_function(camera):
         camera_geometry = camera.geometry
    # do something with camera and camera_geometry
    # this code does not even realize that something was changed above it.
```
This will throw an exception in the users face, but they had enough time to port....

dneise:
fixes #80 if we want that

merging this PR does not change anything for existing code. So it is safe to merge. 
Oh . tests are missing .. damn ... gimme a sec

### DIFF
--- a/digicampipe/utils/tests/test_camera.py
+++ b/digicampipe/utils/tests/test_camera.py
@@ -1,0 +1,30 @@
+'''
+testing digicam.utils.Camera here.
+Not all features it inherits ... just the two new features are tested.
+
+'''
+
+
+def test_can_be_constructed_without_config_path():
+    from digicampipe.utils import Camera
+    cam = Camera()
+
+    # This is calling the __init__ of its base-class. So I do not need to
+    # test that that __init__ really worked.
+    # just some spot checks maybe...
+
+    assert len(cam.Pixels) == 1296
+    assert len(cam.Modules) == 108
+    assert len(cam.Clusters_7) == 432
+    assert len(cam.Clusters_19) == 432
+
+
+def has_attribute_geometry():
+    from digicampipe.utils import Camera
+    cam = Camera()
+
+    g = cam.geometry
+
+    # just some spot checks:
+    assert g.neighbor_matrix.shape == (1296, 1296)
+    assert str(g.pix_x.max()) == '425.25 mm'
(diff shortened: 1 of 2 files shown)


digicampipe/utils/tests/test_camera.py
```diff
@@ -0,0 +1,30 @@
+'''
+testing digicam.utils.Camera here.
+Not all features it inherits ... just the two new features are tested.
+
+'''
+
+
```
calispac: There is definitely a need for more tests regarding `cts.Camera` but this does not belong here !

PR #84, opened 2018-01-18
Console scripts example

dneise:
Example how console scripts would work.

Not too much code is needed for this. 
In order to try it out. you have to "install" digicampipe again, so that setup.py creates command line interfaces.

```
cd to your digicampipe folder
pip install -e .
cd somewhere else ... e.g. where your SST1M data is.
```

then do `digicam_baseline --help`
in case you are like me and have forgotten what options that program needs

then call for example: `digicam_baseline dark.npz SST1M01_20171030.002.fits.fz`
In my case it took ~25sec ... pretty nice.


digicampipe/scripts/baseline.py
```diff
@@ -42,7 +42,7 @@ def main(baseline_file_path, files, unwanted_pixels=[]):
         pass
 
 
-if __name__ == "__main__":
+def entry():
     args = docopt(__doc__)
     args['--unwanted_pixels'] = [
         int(x) for x in args['--unwanted_pixels'].split(',') if x]
```
calispac: Can we still keep the "main" and call `entry()` in it ? that way you can still call it the "regular" way ?

digicampipe/scripts/dark_evaluation.py
```diff
@@ -109,7 +109,7 @@ def main(
     plt.show()
 
 
-if __name__ == "__main__":
+def entry():
     args = docopt(__doc__)
     args['--unwanted_patches'] = [
         int(x) for x in args['--unwanted_patches'].split(',')]
```
calispac: Can we still keep the "main" and call `entry()` in it ? that way you can still call it the "regular" way ?

digicampipe/scripts/pipeline_crab.py
```diff
@@ -151,7 +151,7 @@ def main(args):
             data_stream=data_stream, output_filename=args['--outfile_path'])
 
 
-if __name__ == '__main__':
+def entry():
     args = docopt(__doc__)
     print(args)
     args['--min_photon'] = int(args['--min_photon'])
```
calispac: Can we still keep the "main" and call `entry()` in it ? that way you can still call it the "regular" way ?

dneise:
The "regular" way is in my opinion the "bad" way .. but sure .. technically this is possible. 
It is also kind of clean .. just two more lines at the bottom of each of these files.

PR #85, opened 2018-01-18
Nsb evaluation tidy up

dneise:
Work in progress, do not merge.

So I am keeping on working on the scripts in the root folder. I try to execute them and if they don't work I try to "tidy them up" until they work.

In the process I usually do:
 * pep8
 * delete commented out parts.

Big commented out parts are noisy. The reader is unsure if this is old, and can be ignored, or if it is some kind of optional code, that can be commented in, to have kind of two examples in one.
Or if the script is just not ready and the reader is looking at something which was not finished and is in an experimental stage.

In this process, I might destroy the actually intended functionality. If this is the case, I probably will not realize that, so if reviewers have time. Please review.


calispac:
Are you ready to merge ?
Also how should we proceed for the plots? Should we create a separate script that plots an outputed file ?

PR #86, opened 2018-01-19
Refactoring of baseline computation

dneise:
Work in progress.

I was looking at files in this folder to get a better overview of digicampipe. 
In the process I pep8-ed the files. 

I think `extract_baseline` is quite large and has complex expressions. I would like to understand these expressions better and write them in simpler forms down. So future readers can understand quicker.

In order to make sure I do not break this, while trying to improve readablility, I will implement a test first.
After all chances are high, that a person modifying code she does not understand, might break it.

So next: a test for `extract_baseline` then refactoing of that function


calispac:
This is an ugly part of the code. Probably the ugliest ^^. We need to take care for this. 
Refactoring of the baseline computation would be one of the next tasks.

dneise:
Okay .. started it .. but can you show me how to call this function? 
An example would be good. I do not know what the `calib_container` is

dneise:
Ah .. you self assigned .. I didn't get that. 

Okay I'll do something else

dneise:
@calispac do you still work on this? Or should I give it a shot?

calispac:
No I haven't found time. But you can go on if you want... sorry.

calispac:
Just to give an update: I was waiting for a Processor() interface to come to start with this. But we still have to decide on what to do and not to do. See #119 and #122 
I suppose we can go ahead with simple functions.

dneise:
Nope

PR #89, opened 2018-01-19
add docopt

calispac:
Adding docopt to environment.yml. Which version is needed?
Reported by @sagreus

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -14,6 +14,7 @@ dependencies:
 - cycler=0.10.0
 - dbus=1.10.22
 - decorator=4.1.2
+- docopt
 - expat=2.2.5
 - fontconfig=2.12.4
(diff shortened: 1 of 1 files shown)


dneise:
I have 0.6.2 ... but 0.6 should be fine.
Any version should be fine, I think

dneise:
@sagreus thanks for reporting! :+1

sagreus:
@dneise 
conda version : 4.3.27

dneise:
Ah I see... 
To my understanding that version already has docopt included: 
https://docs.anaconda.com/anaconda/packages/old-pkg-lists/4.3.1/py35

Hm .. I don't get it .. anyway including it in the env.yml is certainly correct ... somehow.

Issue #90, opened 2018-01-19
let digicampipe.event_stream and ctapipe.io.fits.ZFitsFileReader converge


dneise:
So I started a PR in ctapipe, to get **our** protozfits reader into ctapipe. 

Disclaimer:
I stress the word "our" because the protozfits format is of course flexible enough to contain any telescope data, but I only know about SST1M. So this is no "general reader".

The ctapipe PR is here: https://github.com/cta-observatory/ctapipe/pull/598
That PR depends on another PR, which is still not merged: https://github.com/cta-observatory/ctapipe/pull/613

-----

Anyway. Now since the ctapipe starts to converge on a common interface for their `EventFileReader`s, we should use the same interface of course. So let's dream for a moment and assume that ctapipe PR 598 is merged tomorrow. Then of course digicampipe wants to **import** the reader from ctapipe instead of using our own. Since the more we import from ctapipe, the more we are "compatible" with them. Until hopefully one day, most of the "heavy duty code" is moved to ctapipe.

The interface looks like this (basically, not literally):
```python
EventFileReader(input_url="path/to/file.whatever", max_events=None)
```

While our interface is much larger:
```python
def zfits_event_source(
    url,
    max_events=None,
    camera_geometry,
    camera,
    allowed_tels=None,
    expert_mode=False
):
```

So the ctapipe reader cannot do what our reader does, since we cannot give it the necessary input parameters.

Keep in mind, since #81 `camera_geometry` is an attribute of `camera`. So our interface just got smaller. 

### What is missing?

- no `expert_mode`. The ctapipe reader does not know `expert_mode` it always delivers everything, so `expert_mode` is always true. The reason is, that expert mode is not so much slower, and mem is generally not an issue, since we are in a streamed environment. (Only ever 1 event in mem).

 - no `camera` (and no `camera_geometry`). The digicampipe `zfits_event_source` requires two parameters: `camera` and `camera_geometry`. It does not do anything with these parameters but "plugging" them into each event. This has (not yet) anything to do with event reading, so it can be done as well **after** the `event_source`. so e.g. this: 

```python
import digicampipe.utils Camera as DigiCam
digicam = DigiCam()
for event in zfits_event_source(
    url='some/path/to/file.fits.fz', 
    camera=digicam, 
    expert_mode=True
):
    # do stuff with event.
    pass
```

would now become this:
```python
from ctapipe.io.zfits import ZFitsFileReader
import digicampipe.utils Camera as DigiCam
[...]

PR #91, opened 2018-01-19
Use ctapipe zfits reader

dneise:
**Do not merge** .. this cannot work at the moment, since the cta has not merged the PR yet.

This is connected to  #90


dneise:
You see how much code we loose in digicampipe, without loosing functionality. 

One feature got lost, which getting the number of events. In a streamed environment, the number of events is not defined. A stream might be really some socket connection streaming a virtually unlimited amount of events. 

So our code should not need the number of events.

We should train ourselves thinking in event streams not so much in files... might take some time to get used to it.

dneise:
This PR is confusing me .. I will close it until ctapipe is advanced further

Issue #95, opened 2018-01-25
ZFile: `run_id` is always zero


dneise:
Title says it all. The `run_id` is always zero. Should we try to read the run_id from the filename? 

Is it possible that a run has a run_id of zero? If so we should use np.nan or maybe -1 as a marker for "run id is unknown".

We could say: if we can read the run_id from the filename, then we use it, if not we assign -1 or np.nan or None to run_id.

What do you think?

dneise:
I PR #96  @calispac and me discussed about about this issue.
We should have done that here .. sorry

dneise:
The result of the discussion to my knowledge was: 

Yes, `run_id` is always zero and we are happy with that.

PR #96, opened 2018-01-25
ZFile: Do not read RunHeader

### LINKED ISSUE #93: Error: Table RunHeader could not be found in input file
I am running into this issue in some of the scripts. For instance when running "dark_evaluation.py" it throws this message:

Open has failed. Error: Table RunHeader could not be found in input file. Aborting. while reading file digicampipe/tests/resources/example_100_evts.000.fits.fz and table RunHeader
/home/paco/CTA/ctasoft/digicampipe/digicampipe/io/protozfitsreader.py:185: UserWarning: digicampipe/tests/resources/example_100_evts.000.fits.fz has no RunHeader
  "{} has no RunHeader".format(self.fname))

Maybe it is related with some formatting in "protozfitsreader"? 
I am using a fresh installation of digicampipe.

dneise:
This should fix #93.

I removed reading the RunHeader and the RunTails completely from the ZFile class.

An issue which was always there, but only now is clearly visible is the `run_id`. Since we never head a RunHeader, the `run_id` was always zero. 

I've created a new issue #95 to discuss this. It should not have an effect on this PR, since it was always like this.

### DIFF
--- a/digicampipe/io/protozfitsreader.py
+++ b/digicampipe/io/protozfitsreader.py
@@ -173,19 +173,9 @@ class ZFile:
         self.fname = fname
         self.eventnumber = 0
         self.is_events_table_open = False
-
-        try:
-            self.__open_runheader()
-            self.numrows = rawzfitsreader.getNumRows()
-            self.header = L0_pb2.CameraRunHeader()
-            self.header.ParseFromString(rawzfitsreader.readEvent())
-            self.run_id = toNumPyArray(self.header.runNumber)
-        except (ValueError, TypeError):
-            warnings.warn(
-                "{} has no RunHeader".format(self.fname))
-            self.__open_events()
-            self.numrows = rawzfitsreader.getNumRows()
-            self.run_id = 0
+        self.__open_events()
+        self.numrows = rawzfitsreader.getNumRows()
+        self.run_id = 0
 
     def __next__(self):
         if self.eventnumber < self.numrows:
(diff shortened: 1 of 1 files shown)


calispac:
It used to be red from : https://github.com/cta-sst-1m/digicampipe/blob/8a259b63fda035732d112686abba9f192b1a1387/digicampipe/io/zfits.py#L77

sagreus:
It works for me

dneise:
> I am always confused with event_id and run_id... However we use to have a unique identification number for each event. From what I see it is not present anymore, correct?

It was never present. `run_id`was always the same number: zero. Because there was never a RunHeader. This is at least what I understood from Etienne

dneise:
My understanding :

 * "run" is a synonym for "file".
 *  `event_id` is a unique id **inside a run**
 * runs are uniquely identifiable by their filename. 
   Some other experiments uniquely identify runs by a tuple of (date, run_id). 
 * The 3-tuple (date, run_id, event_id) might be used to uniquely id an event of a certain detector.
 * the 4-tuple (tel_id, date, run_id, event_id) might be used uniquely id an event of CTA.
 * Alternatively, the 2-tuple (tel_id, gps_time) might be used uniquely id an event of CTA.

dneise:
This is the oldest version of `ZFile` I could find.
https://github.com/cta-sst-1m/digicampipe/blob/0f9ee1bd681d0f0aeb20ed02b51cdc3f0273a3b1/digicampipe/io/protozfitsreader.py

I also set `run_id` to zero. Since this would have thrown an exception:
https://github.com/cta-sst-1m/digicampipe/blob/0f9ee1bd681d0f0aeb20ed02b51cdc3f0273a3b1/digicampipe/io/protozfitsreader.py#L56-L68

And therefore this always went into the except-block
https://github.com/cta-sst-1m/digicampipe/blob/0f9ee1bd681d0f0aeb20ed02b51cdc3f0273a3b1/digicampipe/io/protozfitsreader.py#L110-L115

However, since the "rule": "Never let an error pass silently" was broken here, one did maybe not realize that the `try` never worked, and **always** the exception block was executed.

calispac:
Ok this PR is not intended to solve the `event_id` issue and it solves the RunHeader issue so we can merge ?

BTW run_id = np.nan or None I think...

Etienne12345:
run_id is indeed always 0. This is because we do not have a master yet, hence the disk-writer knows nothing about runs. I agree that -1 would make more sense: I'll add it to the disk-writer before we take more data.

Issue #94, opened 2018-01-25
confusion: multiple protozfitsreader


dneise:
Since the 1st time I saw digicampipe, it brought its own version of protozfitsreader.py in the io/ folder.

The protozfitsreader package (tar-ball) provided by Etienne also brought a file named protozfitsreader.py. This file was to my knowledge not used by digicampipe. digicampipe used the protozfitsreader-package like this:

```
inside io/protozfitsreader.py:

from protozfitsreader import rawreader
# use rawreader only
```

This multiplicity is confusing, not only to other developers, but it also confuses me, and I've been working on the reader since some time. I can only imagine how confusing this must be to other devs who just started to work with this reader.

Since we have all the Python code now collected inside the cta-sst-1m orga I think we should tidy this up.

Note:

While moving the tar-ball based protozfitsreader package into a git repo, I also modified the structure. protozfitsreader became a package containing 3 python-files and thus the former protozfitsreader.py became `protozfitsreader/__init__.py`. 

### How to tidy this up?

I would like to remove `protozfitsreader/__init__.py`, since it was never used by digicampipe. However I do not know if there are other people using `protozfitsreader/__init__.py`.

Since we have a package called `protozfitsreader`, I think digicampipe should change the naming of `digicampipe/io/protozfitsreader` to something different. 

Looking into `digicampipe/io/protozfitsreader` it contains a class `ZFile`, which spits out `Event`s when iterated over.

So I would propose to rename `digicampipe/io/protozfitsreader` -> `digicampipe/io/zfile`

In `digicampipe/io/zfits.py` we would find this

    from .zfile import ZFile

instead of the current:

    from .protozfitsreader import ZFile

You see how tiny the difference is? 

This would technically work, but import the "wrong" ZFile:

    from protozfitsreader import ZFile

dneise:
Ah another option would be moving `digicampipe/io/protozfitsreader.py` to `protozfitsreader/__init__.py`.

Then the code in `digicampipe/io/zfits.py` would change from:

    from .protozfitsreader import ZFile

to

    from protozfitsreader import ZFile

calispac:
>Ah another option would be moving digicampipe/io/protozfitsreader.py to protozfitsreader/__init__.py.

I agree the protoreader should go in protozfitsreader package. It was a mistake at the start to put it in `digicampipe`.

PR #101, opened 2018-01-25
Move protozfitsreader to its own package

dneise:


dneise:
This can only work when 
https://github.com/cta-sst-1m/protozfitsreader/pull/10
is merged and protozfitsreader v0.43.2 is released.

dneise:
Okay so this is still waiting for this:
https://github.com/cta-sst-1m/protozfitsreader/pull/10

to be merged, which should only be a formality ...

calispac:
what is the status here?

dneise:
This was written against a hypothetical protozfitsreader release v.0.43.2 (which was never created). Instead the features we need here are in the freshly created release v0.44.1

So let me quickly change this...

dneise:
Okay ... now it becomes interesting again .. tests are running ... but if they are passing, we will see. ...

dneise:
This will also fix #94

dneise:
Test are passing, I assume we can merge.

Issue #102, opened 2018-02-05
protozfitsreader issues


dneise:
@calispac found some issues with the current protozfitsreader.py in a PR which wanted to *move* protozfitsreader.py from this repo to the dedicatd protozfitsreader repo. 

https://github.com/cta-sst-1m/protozfitsreader/pull/10

We should tidy this up before we move the code around. So we do not mix tidy-up & move ... and in the end do not know why stuff is not working anymore.

PR #103, opened 2018-02-05
Protozfitsreader issues

dneise:
working on #102


dneise:
@calispac Please have a look at the first commit in this PR. I changed the tests according to your comments in https://github.com/cta-sst-1m/protozfitsreader/pull/10

And now the tests are not working anymore. 

The dtype of baseline as it is read from the protozfitsfile seems to be really a signed integer, not an unsigned one.

dneise:
Opening a new PR without the things that do not work

PR #104, opened 2018-02-05
Protozfitsreader issues

dneise:
working on #102 

which is basically discussed here:
https://github.com/cta-sst-1m/protozfitsreader/pull/10


dneise:
@calispac please have a look at 88d9ab8. In that commit I tried to remove the `self.eventnumber` counter, since we both agreed it is not needed. 

I still agree it is not needed, but the rawreader crashes when one tries to read more events that there are in the file. An exception or an empty string or so would be nicer, c.f.
```
In [1]: from protozfitsreader import rawzfitsreader

In [2]: path = 'digicampipe/tests/resources/example_10_evts.000.fits.fz'

In [3]: rawzfitsreader.open(path+':Events')
Out[3]: 0

In [4]: for i in range(10):
   ...:     string = rawzfitsreader.readEvent()
   ...:     print(i, len(string))
   ...:     
0 171684
1 171684
2 171684
3 171684
4 171684
5 171684
6 171684
7 171684
8 171684
9 171684

In [5]: string = rawzfitsreader.readEvent()
Event number should be between 1 and 10 while you requested event 11
Segmentation fault (core dumped)
```

dneise:
This is now avoiding the crash, without having a counter. 

Now I have to remove the test, which asserts that this counter exists of course.

dneise:
Now I would like to show you something .. One can now make `ZFile` much much simpler .. it is basically just a function with a generator expression inside.

calispac:
If it is simplier with a function then why not... but we should keep in mind ctapipe structure. 
Although I suppose that the function can stay and the ctapipe class would just call this function right ?

also I think `rewind_table()` can be usefull, no?

dneise:
Yeah .. you are right, forget the stuff with the generator function ... that went too far.

dneise:
> also I think rewind_table() can be usefull, no?

any class and its methods we write create interfaces for future developers to come. For them, learning our iterfaces is like leaning a new language. every class is a noun, every function or method is a verb. 

If one can say a thing perfectly with one verb, why should we create multiple verbs for the same thing. 

So the question is not so much: Would some function be possibly useful for somebody, but rather, can somebody achieve the same thing already with another function? Removing duplicate functionalitiy is important.

Here rewinding the table is exactly the same as simply reopening the file, no? So why having two functions, `ZFile.__init__` and `ZFile.rewind_table` to do exactly the same?

It might be they are not doing the same. However `rewind table` is not used anywhere so I as a developer cannot really see the difference yet from studying existing source code, since nobody is using one of the two functions ... So even if `rewind table` is useful. I would have to guess(!) what it does, since no test exists which asserts that my assumptions are correct ... not very helpful is it?

dneise:
Or in other words: Does the Python text file reader `open('path/to/testfile.txt)` have an interface to "rewind"? No ... why not? ... well maybe because it is not needed.

Etienne12345:
Just so that you know, rawzfitsreader does have random access.

dneise:
Ah yes, I remember we talked about that a while ago and I had a look at rawzfitsreader ... and I did not find a way to get a specific event... but I forgot to ask you about it again.
```
In [1]: from protozfitsreader import rawzfitsreader
In [2]: dir(rawzfitsreader)
Out[2]: 
['getNumRows',
 'listAllTables',
 'open',
 'readEvent',
 'rewindTable']
```

Typically one can execute `help()` on a function or class and fine some stuff our about it. but e.g. `help(rawzfitsreader.getNumRows)` gives:
```
Help on built-in function getNumRows in module protozfitsreader.rawzfitsreader:

getNumRows(...)
    Get number of rows / message in table
(END)
```

So I had a look into the code I found ... and tried a bit out ... I came up with this:
```
path = 'digicampipe/tests/resources/example_10_evts.000.fits.fz'
```
----
### listAllTables
```
In [5]: rawzfitsreader.listAllTables(path)
Attempting to list all tables from digicampipe/tests/resources/example_10_evts.000.fits.fz
Out[5]: 'Events:'
```

So listAllTables takes a path. What if the file does not exist? It prints debug log to stdout, might turn out to be a bit noisy in an application, which checks 1000 files for their tables ... but not important here.
```
In [6]: rawzfitsreader.listAllTables('foo.bar')
Attempting to list all tables from foo.bar
Cannot open foo.bar
Out[6]: ''
```
Empty string .. no file not found Exception .. but useful anyway.

----
### getNumRows
```
In [8]: path
Out[8]: 'digicampipe/tests/resources/example_10_evts.000.fits.fz'

In [9]: rawzfitsreader.getNumRows(path)
Out[9]: 0
```

Hm ... strange ... I was sure the result should be 10 ... not 0.

What if the path does not exist?
```
In [10]: rawzfitsreader.getNumRows('foo.bar')
Out[10]: 0
```

Aha .. also zero ... 

I did not understand this interface, but studies the code .. and found this function does not take an argument. The correct usage is like this:
```
In [13]: rawzfitsreader.open(path)
Out[13]: 0

In [14]: rawzfitsreader.getNumRows('foo')
Out[14]: 10
```

So first one opens the file, then one can check for the number of rows. One can give any argument to `getNumRows` it does not care. 

-----

###open

open we have already seen above, it takes one string argument, and one can add the name of the table with a colon, like this:
```rawzfitsreader.open("some/path/to/file.zfits:Events")```

So that is clear ... 
Also, open with a wrong path returns -1.
```
In [16]: rawzfitsreader.open('foo')
Open has failed. Error: Key 'PBFHEAD' not found. while reading file foo and table 
Out[16]: -1
```

very good.

----

###rewindTable

Next I tried rewindTable, I think. 
```
In [18]: rawzfitsreader.rewindTable()
---------------------------------------------------------------------------
SystemError                               Traceback (most recent call last)
<ipython-input-18-a541eea13629> in <module>()
----> 1 rawzfitsreader.rewindTable()

SystemError: <built-in function rewindTable> returned NULL without setting an error
```

I tried again, open the file ... read at least one event ... try to rewind ... with and without path as argument:
```
In [26]: rawzfitsreader.open(path)
Out[26]: 0

In [27]: s = rawzfitsreader.readEvent()

In [28]: len(s)
Out[28]: 171684

In [29]: rawzfitsreader.rewindTable(path)
---------------------------------------------------------------------------
SystemError                               Traceback (most recent call last)
<ipython-input-29-fde7f2ca4b97> in <module>()
----> 1 rawzfitsreader.rewindTable(path)

SystemError: <built-in function rewindTable> returned NULL without setting an error
```

No idea what I am doing wrong here.

----

### readEvent

Well this I have seen in code already, no path argument needed, just `open` the file and read the event, which is a string, like above:
```
In [26]: rawzfitsreader.open(path)
Out[26]: 0

In [27]: s = rawzfitsreader.readEvent()

In [28]: len(s)
Out[28]: 171684
```

And this can the interpreted as a `CameraEvent` like this:
```
In [35]: from protozfitsreader.L0_pb2 import CameraEvent

In [36]: e = CameraEvent()

In [37]: e.ParseFromString(s)
Out[37]: 171684
```

However how random event access works, I have not found out and also not found an example for it.

File: digicampipe/io/zfits.py [hunk 1]
```diff
@@ -47,11 +47,11 @@ def zfits_event_source(
...
     loaded_telescopes = []
 
-    for event in protozfitsreader.ZFile(url):
-        if max_events is not None and event.event_id > max_events:
+    for event_counter, event in enumerate(protozfitsreader.ZFile(url)):
+        if max_events is not None and event_counter > max_events:
             break
 
-        data.r0.event_id = event.event_id
+        data.r0.event_id = event_counter
```
calispac: `data.r0.event_id = event.event_number` ?

dneise: ???

calispac: `event_counter` is taken from the `enumerate()` so we don't need to store it as we said.
If I look at the `Event` class then the only attribute I find close to a unique identifier is `event.event_number`

dneise: Yes, but see below: `event.event_number` is already assigned to `data.r0.tel[tel_id].camera_event_number`.
We do not want to duplicate this do we?

calispac: ok did not see it... sorry.
So I think we should get rid of those counters in the future. Naming should be : event_id for unique identification number and event_counter if needed somewhere.

dneise: Actually this is totally messed up:
```python
class R0Container(Container):
    event_id = Field(-1, "event id number")
    tel = Field(Map(R0CameraContainer), "map of tel_id to R0CameraContainer")

class R0CameraContainer(Container):
    camera_event_id = Field(int, 'Camera event number')
    camera_event_number = Field(int, "camera event number")
```

3 event ids ... or numbers ...

dneise:
What happened now?

digicampipe/io/zfits.py
calispac: I think it is supposed to be like this

```python
class R0Container(Container):
    array_event_id = Field(-1, "Array event identification number")
    tel = Field(Map(R0CameraContainer), "map of tel_id to R0CameraContainer")

class R0CameraContainer(Container):
    camera_event_id = Field(int, 'Camera event identification number')
```

calispac: or maybe using the same naming for both array and telescope level is better. 

that way you can do : 

`data.r0.event_id`

and

`data.r0.tel[x].event_id`

Etienne12345: Looked at the code, indeed the random access was not interfaced to python. So it should only be readEvent() and rewindTable(). 

I also fixed the crash after reading to many events, but I am not sure what should be the behavior there:
- looks like I can raise a python exception from C++
- or I can just return an integer object instead of a protobuf object

calispac: Thanks!

Where is the random access missing actually? Are we supposed to call a function existing in `rawzfitsreader` or do you miss it there?

PR #105, opened 2018-02-09
Slow data

yrenier:
Automatic generation of slow data container based on fits files.
Special comments in headers allows to set the description of the containers.
Synchronization with digicam data done on the 'timestamp' property of each slow data.

Tests are available, but as the data files are pretty big (>100 Mo) , I did not include them in the resources folder yet. We probably need something to make example files, containing only few events.


yrenier:
@pep8speaks

dneise:
we've realized, pep8speaks will update its report every now and then ... so there is in principle no need to delete the comment of it ... it will shrink and shrink and shrink .. until you get some glasses of beet :-D

yrenier:
ok thanks for the tip

dneise:
No prob ... I don't know how often it updates ...

digicampipe/io/containers.py
```diff
@@ -1,39 +1,46 @@
...
+    'R1CameraContainer',
+    'DL0Container',
+    'DL0CameraContainer',
+    'DL1Container',
+    'DL1CameraContainer',
+    'ReconstructedContainer',
+    'ReconstructedShowerContainer',
+    'ReconstructedEnergyContainer',
+    'ParticleClassificationContainer',
+    'DataContainer'
```
calispac: Missing `SlowDataContainer`

digicampipe/io/event_stream.py
```diff
@@ -1,25 +1,112 @@
...
+    files = [info_struct['hdu']._file.name for info_struct in info_struct_list]
+    ts_min_all = np.array(ts_min_all)
+    ts_max_all = np.array(ts_max_all)
+    files = np.array(files)
+    indexes_slow_file = np.logical_and(
+        ts_min_all < data_ts,
+        ts_max_all > data_ts
+    )
+    if np.sum(indexes_slow_file) == 0:
+        # print('ERROR: no slow file found !')
```
calispac: The print/warning should be restored I think. Otherwise there is no way to see want went wrong?

digicampipe/io/generate_slow_containers.py
```diff
@@ -0,0 +1,203 @@
...
+container and a function to fill it is generated.
+Synchronisation of slow data with digicam data is made acording to the timestamp
+property of the slow data.
+In case you want to use slow data files with a different format than the one 
+currently used, empty the digicampipe/tests/resources/slow/ folder, place the 
+slow data you want to use in it an run this script.
+"""
+
+
+def unit_parser(column):
```
calispac: Maybe we could make good use of `astropy.units` here. But this requires some major change in `digicampipe`

digicampipe/io/generate_slow_containers.py
```diff
@@ -0,0 +1,203 @@
...
+            if str(values['units']) not in ['none', 'None', None]:
+                data_printed = (key, values['type'],
+                                values['desc'], values['units'])
+                print('    %s = Field(%s, "%s", unit=u.%s)'
+                      % data_printed, file=file)
+            else:
+                print('    %s = Field(%s, "%s")'
+                      % (key, values['type'], values['desc']), file=file)
+    print('\n', file=file)
+    return class_name
```
calispac: Sorry I don't understand why this function does : 
1. read class_name
2. do bunch of stuff with hdu
3. return class_name

Should it not return the container ?

calispac: ah I see... the function should write some data into a file write ? If so I suggest renaming `generate_` to `write_` or something else. generate indicates (to my opinion) that it creates the object (inside memory)

calispac: Also file=sys.stdout as default value is maybe miss leading no? because in which situation would you use that except  for debugging?

digicampipe/io/containers.py
calispac: sorry it is correct...

digicampipe/io/generate_slow_containers.py
yrenier: I do use astropy.units (see generate_container_class() ), that parser is just to return units compatible with astropy.units when the description in the fits file is not.

digicampipe/io/generate_slow_containers.py
yrenier: the goal of the generate_... functions is to create parts of the slow_container.py file.
It is indeed defaulted to sys.stdout for debug (and to avoid to recreate the slow_container.py file accidentally).
I can change the default value if you want.

digicampipe/io/event_stream.py
yrenier: the warning is issued in add_slow_data(), that why I disabled it here.

digicampipe/io/generate_slow_containers.py
calispac: ah yes it creates units via Field()... sorry

digicampipe/io/generate_slow_containers.py
calispac: ok. then I think having no default value is better...
because the user would just run the function and wonder what it does except just printing stuff. 
Forcing the user to choose a file also diminishes the risk of overwriting, no?
As you prefer.

digicampipe/io/generate_slow_containers.py
```diff
@@ -0,0 +1,203 @@
...
+            'digicampipe',
+            'tests/resources/slow/*.fits'
+        )
+    )
+
+    # get list of files to treat
+    # (to not treat classes more than once).
+    files_to_process = []
+    classes_to_process = []
+    for path_file in filenames:
```
calispac: It would be nice to have this part of the code to be handled in one function/class. Avoiding that the user does it in the wrong order. someting like:

```python
def write_slow_control_data_container(example_slow_file, file_to_write):
    ...
```

dneise:
I think it would indeed be good to have a couple of example files, so we can test this code (either on travis or on our own machines).

dneise:
Or it might be good if somebody could tell me where to get some slowdata files

yrenier:
the files I used in test_slow_data.py are:

slow data files (from /home/ctauser/data/aux/2017/12/4/SST1M01/ on sst1mserver) : 
DriveSystem_20171204_000.fits 
MasterSST1M_20171204_000.fits 
PDPSlowControl_20171204_000.fits
PDPSlowControl_20171204_001.fits
SafetyPLC_20171204_000.fits

digicam data file (from /home/ctauser/data/raw/2017/12/04/SST1M01/ on sst1mserver) :
SST1M01_0_000.006.fits.fz

Just put this files in test/ressources/slow and it should almost work.
I say almost as it seems to be something wrong with the PDPSlowControl clock, so no synchronized
data is found for that type of file.

dneise:
Unfortunately I have still no access to the sst1mserver. 
I have only the files I can download with the script from Etienne, which was mentioned in the analysis call on tuesdays.

digicampipe/tests/test_slow_data.py
```diff
@@ -0,0 +1,135 @@
...
+    plt.figure()
+    plt.subplot(2, 1, 1)
+    plt.plot(ts_slow - np.min(ts_data), ts_data - np.min(ts_data), '.')
+    plt.xlabel('ts(slow data) - ts_min, s')
+    plt.ylabel('ts(digicam) - ts_min, s')
+    plt.subplot(2, 1, 2)
+    plt.plot(diff, '.')
+    plt.xlabel('event')
+    plt.ylabel('ts(digicam) - ts(slow data), s')
+    plt.show()
```
dneise: I assume this is a "test-plot", which a developer needs to look at and decided if the code is behaving correct?
Would it be possible to convert this test into an `assert`ion, so it can be tested all the time by travis?

Etienne12345:
Do you still have an account at ISDC ?
If so, everything is in on the cluster, at /nasAcademique/archive/walter/sst1m/

dneise:
> Do you still have an account at ISDC ?
If so, everything is in on the cluster, at /nasAcademique/archive/walter/sst1m/

Nice ...thanks a lot!
Is this somehow mirrored on a regular basis or more like a snapshot?

Etienne12345:
On a regular basis, similarily to the fact archive. Only the transfer process is stopped right now because the telescope is somewhat "off" and I must change the pipeline to take the data from UNIGE rather than IFJ these days.

dneise:
Awesome

digicampipe/tests/test_slow_data.py
yrenier: It is indeed a test plot to check the synchronization of slow data with digicam data, but the test corresponding to that plot is already implemented : 
    assert(all(np.array(diff) <= 1.1))

dneise:
This automatic code generation is quite cumbersome. In principle it requires a two level review, not only the code which is created needs a review ... also the code which generates code.

Are we sure this is needed? 

If code generation is needed, maybe some templating engine would be nicer here? like jinja2 or so?
This would make code generation a little bit nicer to read.

for example this
```python
def write_fill_slow(file_names, file):
    print('def fill_slow(class_name, event, hdu, slow_event):', file=file)
    first = True
    for path_file in file_names:
        class_name = get_class_name(path_file)
        if first:
            print('    if class_name == "%s":' % class_name, file=file)
            first = False
        else:
            print('    elif class_name == "%s":' % class_name, file=file)
        print('        event = fill_%s(event, hdu, slow_event)'
              % class_name, file=file)
    print('    else:', file=file)
    print('        print("ERROR in fill_slow():'
          ' class %s not known." % class_name)', file=file)
    print('        print("Try to regenerate slow data containers ?")',
          file=file)
    print('    return event', file=file)
```

can be replaced by this:
```python
def write_fill_slow_2(file_names, file):
    from jinja2 import Template
    class_names = [get_class_name(path) for path in file_names]

    t = Template('''
def fill_slow(class_name, event, hdu, slow_event):
    fillers = {
        {% for cn in class_names %}
        {{cn}}:fill_{{cn}},
        {% endfor %}
    }
    event = fillers[class_name](event, hdu, slow_event)
    return event
''')
    print(t.render(class_names=class_names), file=file)
```

dneise:
I will make a proposal how this code would look using jinja  ... so we can have a look and decide

yrenier:
I did not know jinja2, some reading for me :-)

dneise:
Ah jinja2 is only one of many templating engines .. I do not even know if there are better ones.

dneise:
Ah apart from that, many of the code generating functions do both:
 * generate the code **and**
 * print it

one could also split this, so that most functions return strings ... and then writing the stuff to file or to stdout or whatever is done somewhere else.

dneise:
Ah and a general question ... I see this is creating many of these ctapipe Containers .. I still do not know how ctapipe will handle slow data. 

So are we somehow sure this is the right way to go?

yrenier:
The automatic code generation was wanted because we expect the slow data files to change significantly.
Using automatic code generation would allow to read old files or update the current version of the container just by re-running the generating part.

yrenier:
Maybe we should discuss this directly at the meeting tomorrow ?

dneise:
Sure, we can ... but I'd like to prepare myself a bit ... I admit have not found much information about slowdata in ctapipe, so I just wanted to ask, if you have some docu or some discussions in a ctapipe issue or so, I could read before tomorrow? 

Just so we are somehow on the same page?

yrenier:
I saw nothing about slow data in ctapipe when I checked in December and I must admit I did not check since.

dneise:
Ah .. I've just seen this:
```python
def get_class_name(path):
    filename = os.path.basename(path)
    m = re.match('(?:slow_)?(?P<class>[\w]+?)_[\d_]+\.fits', filename)
    class_name = m.group("class")
    return class_name
```

can also be read from the fits header like this:
```python
def get_class_name(path):
    return fits.open(path)[1].header['SUBSYS']
```

reading it from the header takes ~2ms and I assume getting the class_name from the filename might be quicker. I was just wondering if filename and fits header will always be consistent?

dneise:
@Etienne12345 
I am unable to find this file: SafetyPLC_20171204_000.fits
at ISDC. in the place where this file should be is a folder named SafetyPLC/ with a lot of .log files ...

yrenier:
If the class name is in the header, that's indeed much better to use that instead of something extracted form the filename !

dneise:
OMG!
 there is a "bug" in pytest ... that means, I misunderstood how "skipping" tests in pytest works.

Remember I marked some tests with `@pytest.mark.slow` and by default we do not execute these tests, because in the `pytest.ini` the line:

    -k-slow

leads to pytest by default skipping these tests? (`k` means s**k**ip) 

----

Well for some reasons, tests which are in a file containing the string "slow" or which are in a function with "slow" in the name are also skipped.

dneise:
So when I execute: simply `pytest` all tests regarding slowcontrol stuff are not even executed.

Etienne12345:
I just copied the file to ISDC by hand (to the SST1M01 directory). 
Looks like the pipeline missed it, must investigate why exactly...

dneise:
great thanks

dneise:
Ah .. the tests are nearly passing on my machine now:

Just one thing:
```
example_file_path = resource_filename(
    'digicampipe',
    os.path.join(
        'tests',
        'resources',
        'slow',
        'SST1M01_0_000.006.fits.fz'
    )
)
```

This file "SST1M01_0_000.006.fits.fz" I do not have. I replaced that path with: "SST1M01_20171204.006.fits.fz" then it worked.

Etienne12345:
Isn't this test supposed to use the small example file that is provided along with the package ? 

Nice to read the the tests pass ! :)

dneise:
> Isn't this test supposed to use the small example file that is provided along with the package ?

I totally agree, but there are apparently no small slow-data-files, that go along with the 10evts or 100 events data file. I think we need your help to make such files.

Etienne12345:
Ah right, I see. Let's discuss this during tomorrow's telconf: I have a general idea about who to produce this data, but it will take me some time before I can work on this.

yrenier:
@Etienne12345 if you could provide tomorrow a small tutorial on how to make such files, I would be happy to create them.

Etienne12345:
That's the thing: I have the general idea, but not a tutorial yet... Let's discuss tomorrow !

yrenier:
slow data are currently only saved every second, so the tests file should indeed be few seconds long.

dneise:
As a little test I downloaded now the slow-data from the day when the 100event test file was created: 20171031. 

The tests do not pass anymore. I get a lot of these:
no PDPSlowControl data found
no MasterSST1M data found
no DriveSystem data found
no SafetyPLC data found
no DigicamSlowControl data found

Do you think the slow data format was different back then? 

If so, I think I have to regenerate this code. Can you remind me how to regenerate this code?

yrenier:
The timestamp of the slow data much match the timestamp to not get these errors and pass the tests.

To regenerate the slow_containers.py file, just go into the digicam/io folder and do :
`python generate_slow_containers.py`

dneise:
> The timestamp of the slow data must match the timestamp to not get these errors and pass the tests.

I assume there is some kind of window? I mean the timestamp is ns precision or so ... two timestamps can hardly be equal down to the ns.

Etienne12345:
For the slow control, and if I understood correctly that the example events come from one night during which we looked at Crab, there is one night (most probably the first, but cannot be 100% positive) when the timestamps are all wrong in the raw data. Reason was that digicam / camera server did not synchronise, hence digicam clock counter was written to the data rather than Unix timestamps (basically the epoch is reset several times during the night). Do do the analysis I had to estimate the unix time corresponding to the clock reset, and update the timestamps accordingly. Cannot recall either if I had pushed the fixes to the raw data file or if it remains in the analysis only. Let's discuss all this tomorrow.

yrenier:
To be more precise:
for each event  of the digicam data, the digicam timestamp must be between the first and last timestamp in at least one of the slow data files to not get the error messages you see.
Then for each digicam event, the latest slow data with a timestamp smaller than the digicam timestamp is used.

dneise:
I had a look at the slow-data-files I found at ISDC for 20171031.
```
slow/DigicamSlowControl_20171031_000.fits......... 31 12:09:23 31 15:18:18 : 3.1hours
slow/DriveSystem_20171031_000.fits................ 31 12:08:18 31 12:09:16 : 0.0hours
slow/DriveSystem_20171031_001.fits................ 31 12:09:23 31 12:09:49 : 0.0hours
slow/DriveSystem_20171031_002.fits................ 31 12:10:15 31 12:11:24 : 0.0hours
slow/DriveSystem_20171031_003.fits................ 31 12:11:31 31 14:59:19 : 2.8hours
slow/MasterSST1M_20171031_000.fits................ 31 12:08:17 31 14:44:28 : 2.6hours
slow/MasterSST1M_20171031_001.fits................ 31 14:44:49 31 15:17:48 : 0.5hours
slow/MasterSST1M_20171031_002.fits................ 31 15:43:01 01 12:00:09 : 20.3hours
slow/SafetyPLC_20171031_000.fits.................. 31 12:08:18 01 12:00:10 : 23.9hours
```

I learned that: 
 * sometimes files stop after some time .. and new file starts (maybe DAQ restart or so)
 * for some slow-data there are only data from ~12:00 until ~15:00h. 

The 100events example file looked like it was taken in the middle of the night.

dneise:
I think, it would be good, if we create a new ~100events example file, from a day/night where we have also slow-data overlapping with the raw-data.

dneise:
> I had a look at the slow-data-files I found at ISDC for 20171031.

I am stupid .. I was looking at the wrong day .. I have to look at the night: 20171030 .. since the 100event example file was taking in that night ... just early morning 20171031. Stupid.

dneise:
Okay now I got the slow-data files from the correct date: 20171030. 
I re-executed the tests.... did not work. some fields were missing in the slow-data files.
So I re-generated the code ... worked. 
The I re-executed the tests ... did not work. but for a different reason:

test_get_slow_event FAILED

```
    def test_get_slow_event():
        slow_info_structs = get_slow_data_info(slow_file_list)
        for class_name in SLOW_CLASSES:
            ts_min_all = [info_file['ts_min']
                          for info_file in slow_info_structs[class_name]]
            ts_max_all = [info_file['ts_max']
                          for info_file in slow_info_structs[class_name]]
            ts_min_all = np.array(ts_min_all)
            ts_max_all = np.array(ts_max_all)
            for file in range(len(slow_info_structs[class_name])):
                info_struct = slow_info_structs[class_name][file]
                ts_min = info_struct['ts_min']
                ts_max = info_struct['ts_max']
                if ts_min == ts_max:
                    continue
                data_ts_ok = (ts_max + ts_min) / 2
                e, h = get_slow_event(slow_info_structs[class_name], data_ts_ok)
>               assert(e is not None)
E               assert None is not None

digicampipe/tests/test_slow_data.py:72: AssertionError
------------------------------------------------ Captured stdout call -------------------------------------------------
ERROR: several slow files found ! ['/home/dneise/sst/digicampipe/digicampipe/tests/resources/slow/DigicamSlowControl_20171030_003.fits'
 '/home/dneise/sst/digicampipe/digicampipe/tests/resources/slow/DigicamSlowControl_20171030_002.fits']
```

Now my day is over :-D .. have a good evening everybody.

digicampipe/tests/test_slow_data.py
dneise: Ah so this plot can be removed, right?

yrenier:
I got the same problem with multiple files containing the same timestamps. I guess it is related to some issues such as the one Etienne mentioned a few post earlier.
As matter of fact the time of the slow data files are mixed:
-rw-r--r-- 1 503 504  13835520 Oct 30 16:18 DigicamSlowControl_20171030_002.fits
-rw-r--r-- 1 503 504    311040 Oct 30 15:52 DigicamSlowControl_20171030_003.fits
-rw-r--r-- 1 503 504   2911680 Oct 30 16:17 DigicamSlowControl_20171030_004.fits
To get it to work for that night, I had to manually delete some of the conflicting slow data files.
Not so nice, but I believe this clock synchronization problem is now fixed.

Etienne12345:
It looks like this is a separate issue: the clock sync problem I was talking about earlier was between Digicam and Camera server, and thus affecting only the raw data files. 
For the slow control files, I am not aware of a similar problem: let's ask Vitalii during the telconf

dneise:
I think a decision has to made, how to procede with the slow data.
If we want to vote, I would vote for the alternative implementation.

dneise:
Okay, then ... should we merge something until the slowdata are fixed? So people who are interested, can plot stuff?

Or close both PRs since we need something else anyway? 

Or just wait?

yrenier:
you are right, lets merge the alternate then.

dneise:
Closing this. But do not delete branch, might be needed later

PR #107, opened 2018-02-12
rename pytest skip marker from slow to deselect

dneise:
I had some issues with pytest. 

Since we had `-k-slow` in the pytest ini, no test function containing "slow" in its name would be executed. 

I would like to rename this to "deselect" hoping that we simply use the word "deselect" less often in normal function names.


dneise:
I think this one here should be merged next, if somebody has time to quickly review it

PR #108, opened 2018-02-13
Slowdata alternative

dneise:
I was curious how the slowdata would look without the code generation.

I quickly hacked something together. Maybe @yrenier or others would like to have a look before our telcon today.


dneise:
Wait .. I made a mistake .. this branch is somehow based on yves branch .. I'll try to tidy this up ..

PR #109, opened 2018-02-13
Slow data Alternative

dneise:
Hello,
I was curious how the slowdata code would look without the code generation.
So I quickly hacked something.

Maybe somebody like @yrenier or @calispac or @Etienne12345  would like to look at it before our meeting today.

[Example Notebook](https://github.com/cta-sst-1m/digicampipe/blob/slow_data_minimal/how_to_play_with_slow_data.ipynb)


dneise:
I have merged the branch `deselect_tests` from PR #107 into this, because I needed that. Otherwise the tests containing the name "slow" would not even run

dneise:
I am using the 100 events example file for the test, and have created ~150kB of example slow data files. These are small enough for us to store them in the git repo as well.

yrenier:
I think you forgot to include the AuxService.py file.

dneise:
The main new thing is an `AuxService`. It works basically like this:
```python
from digicampipe.io.auxservice import AuxService
drive_service = AuxService('DriveSystem', basepath='/nasAcademique/archive/walter/sst1m')
```
So one makes a service and tells it, which slow data one is interested in and where to find them.
Next, one asks the service for a `row` at a certain time, e.g. like this:
```python
row = drive_service.at(event.r0.tel[1].local_camera_clock)
```

The `row` is at the moment `numpy.recarray`, which can nicely be used to access its attributes using the simple dot notation, like `row.timestamp` or so. It prints itself not too beautiful, but okay-ish.

dneise:
The main code, in the `AuxService` deals with stacking together multiple slow data files per day. 
E.g. for the day 20171030 this is the original slow data folder:
```
DigicamSlowControl_20171030_000.fits  MasterSST1M_20171030_007.fits
DigicamSlowControl_20171030_001.fits  MasterSST1M_20171030_008.fits
DigicamSlowControl_20171030_002.fits  MasterSST1M_20171030_009.fits
DigicamSlowControl_20171030_003.fits  MasterSST1M_20171030_010.fits
DigicamSlowControl_20171030_004.fits  PDPSlowControl_20171030_000.fits
DigicamSlowControl_20171030_005.fits  PDPSlowControl_20171030_001.fits
DigicamSlowControl_20171030_006.fits  PDPSlowControl_20171030_002.fits
DigicamSlowControl_20171030_007.fits  PDPSlowControl_20171030_003.fits
DigicamSlowControl_20171030_008.fits  PDPSlowControl_20171030_004.fits
DigicamSlowControl_20171030_009.fits  PDPSlowControl_20171030_005.fits
DigicamSlowControl_20171030_010.fits  PDPSlowControl_20171030_006.fits
DigicamSlowControl_20171030_011.fits  SafetyPLC_20171030_000.fits
DriveSystem_20171030_000.fits         SafetyPLC_20171030_001.fits
DriveSystem_20171030_001.fits         SafetyPLC_20171030_002.fits
DriveSystem_20171030_002.fits         SafetyPLC_20171030_003.fits
DriveSystem_20171030_003.fits         SafetyPLC_20171030_004.fits
DriveSystem_20171030_004.fits         SafetyPLC_20171030_005.fits
DriveSystem_20171030_005.fits         SafetyPLC_20171030_006.fits
DriveSystem_20171030_006.fits         SafetyPLC_20171030_007.fits
MasterSST1M_20171030_002.fits         SafetyPLC_20171030_008.fits
MasterSST1M_20171030_003.fits         SafetyPLC_20171030_009.fits
MasterSST1M_20171030_004.fits         SafetyPLC_20171030_010.fits
MasterSST1M_20171030_005.fits         SafetyPLC_20171030_011.fits
MasterSST1M_20171030_006.fits
```

The `AuxService` simply reads all slow-data files for this day and concatenates them into a single table.

dneise:
One possible problem with concatenating multiple files is of course when the files have different keys (or column names). Since the code cannot **guess** which column was renamed in what way.

By chance this problem also occured on the day 20171030. 
Some files hat a column called: `timestamp`. while in others the column was called `TIMESTAMP`.
Also some files had a column called `TIME` while others did not have this column.

For the drive files it got even more interesting: there was one files, which contained about 6 columns more than the others... However, this is not a problem. The AuxService does not care about any other column than the `timestamp`, which it needs to look for the event time. 
All other columns are just seen as payload and finally delivered as fields of the `row`.

dneise:
> If I get it right, you generate the slow data container on the fly by reading the slow data ?

Well .. no ... I tried a different approach. I did not create any `Container` class and also non of their instances. 

The Container as it is defined by ctapipe, can not and does not enforce all fields to bw `Containers` themselves. One can put anything into a container. 

So in this approach, I decided to try how the code looks and feels, when I deliberately do not use Containers.

dneise:
You said somewhere in PR #105 that you went for code generation, because you anticipated, that the format of the slow-data files would often. 

If I understood the goal of the `ctapipe.Container` correctly, then it was made, so that people who *read stuff from a container* can trust, that the names of the fields will never change. 

Now if we know the field names will change often, and therefore generate Containers dynamically, we broke this contract anyway. So I saw no point in generating this much code, which does not help at all.

Etienne12345:
Just a quick remark: the multiple files per nights is supposed to happen only when something changed during the night, ie. the output format or number of columns. 
We have many right now because it was commissioning, and a lot of the fine-tuning work happened during the nights. 
In the future, we should expect only one file per night (hopefully). 
Of course it is better to have a reader that can do more ! 
All of the above is from the top of my head: Vitalii can confirm / provide further details this afternoon

dneise:
>  In the future, we should expect only one file per night (hopefully).

Yes, I assumed exactly this. But I also assume, that there might be people which are interested in reading events exactly from such nights. So I was trying to get some slow-data into their hands.

dneise:
Of course there is a way around the Table "merging" problem. 
Simply do not merge them. 

When you get an event, simply open only the right file... and return a row only from that file.

I did not go that way because, then I'd need to keep a table of several files .. too much hassle for my little mind.

And ultimately, when there is only one file per night, we do not have to think about this question anyway. So ...

dneise:
I've added a notebook, which shows how to play with the slow data.

https://github.com/cta-sst-1m/digicampipe/blob/slow_data_minimal/how_to_play_with_slow_data.ipynb

I think it shows nicely some of the problems this implementation still has. (There might be more fundamental problems as well, which are not shown there, since I do not know about them).

Please have a look

dneise:
Okay .. I modified it, so that everything is returned as a namedtuple. I think they are nicer to work with than numpy.recarray rows. I mean in principle numpy.recarrays are nice, but single rows out of a recarray do not act intuitively, c.f. scalar-arrays.

I also removed some code, which was not needed, and actually not wanted.

I also broke out some functions and added some docstrings.

yrenier:
looking at the notebook, the single data are not array anymore which is way more intuitive than the strange array you had before.
However I don t understand what is the problem with the multiple events ?

dneise:
Please see issue #111 I have just opened for this.
It has nothing to do with slow data ... I just realized it now.

dneise:
Independent of #111. What do you guys think? Should we go more in this direction, or more in the direction of #105 or some combination of both?

Python is a dynamic language, classes can be created at runtime. There is no need to generate a python file, then store it, then import it and then use the classes defined in the file. One can directly create the classes and use them in one go. This way the "generate step" would fall away.

But I personally think that this kind of dynamic class creation is complicated, be it via explicit code generation (which at least can be looked at) ... or immediately in memory at runtime.

PR #112, opened 2018-02-14
Quicker event access

dneise:
I propose to make event access easier.

This is  digicampipe, so the default case for all of us is using the digicam. The `event_source` should know about this default and act accordingly. 

People can still create their own camera config files, and create their own camera, like:
```python
my_cam = utils.Camera('my_config_file.txt')
for event in event_stream('event_file.fits.fz', camera=my_cam):
    pass
```

But the default is to use `utils.DigiCam`.

This should make the life of normal people easier, while camera specialists, who want to use several different cameras for testing, need to write a bit more code, but still can do their job.

-----

In addition: I added one "nice-to-have" feature, which is you can either give a list of paths, or a single path, and both works... I so often made the mistake of giving a single file only ... I thought, maybe others do to.

For a comparison of the way it was before, and the way it is now, see: 
https://github.com/cta-sst-1m/digicampipe/blob/nice_event_source_and_pep8/digicampipe/tests/test_event_stream.py


digicampipe/io/event_stream.py
```diff
@@ -1,25 +1,30 @@
...
-
+def event_stream(
+    filelist_or_url,
+    camera=None,
+    camera_geometry=None,
+    expert_mode=None,
+    max_events=None,
+    mc=False
+):
+    if isinstance(filelist_or_url, (str, bytes)):
```
calispac: FYI there is something similar for numpy array and single values : 

https://docs.scipy.org/doc/numpy/reference/generated/numpy.atleast_1d.html

I was quickly looking if there is something similar for strings and lists of strings but could not find it...

dneise: Yes, this is an ugly point, it makes the interface "unclear". 
A function should have a clear interface. 

But, I thought this function is very exposed to the "end-user" so adding this feature felt "okay-ish" ... 

I'm not sure anymore though... maybe its fine

calispac: Yes maybe a bit "hard" to read but I suppose a comment would make it easier

calispac: also you can overwrite the variables right? 

just removing the else statement and: 

```python
def event_stream(
    filelist,
   ....
   )
...
   if isinstance(filelist, (str, bytes)):
      filelist = [filelist]

...
```

dneise: :-D ... I guess you know my opinion about comments. When they are needed, the interface is broken :-D

dneise:
thanks

PR #113, opened 2018-02-14
add __version__

dneise:
now people are able to:
```python
>>> import digicampipe
>>> print(digicampipe.__version__)  # --> 0.1.3
```

This way users can quicker find out if an issue might be due to having an older version installed.

We should make a habit of increasing the lowest version number with every commit.
There is in principle nothing wrong with having two digit minor versions, like: 0.1.25.

When to increase to 0.2 is basically up to personal taste ...


calispac:
thanks nice feature!

PR #114, opened 2018-02-15
Nice event source and pep8

dneise:
WIP


baseline_stability.py
```diff
@@ -14,55 +13,61 @@
...
+        file_list=file_list,
+        camera=DigiCam
+    )
 
     for i, data in zip(range(n_events), data_stream):
-
         for tel_id in data.r0.tels_with_data:
-
-            digicam_baseline[..., i] = data.r0.tel[tel_id].digicam_baseline
+            baseline[..., i] = data.r0.tel[tel_id].baseline
```
calispac: I think here it should access: `data.r0.tel[tel_id].digicam_baseline` meaning it is the baseline computed by DigiCam FPGA.
or it was renamed...

dneise: oh .. yeah that was a mistake! Good catch!

I thought the variable name is too long and I can make it shorter, since in the whole script there is only one "digigcam_baseline", so I thought there is not need for the extra qualification "digicam" ... since we all know we are working on digicam here... 

anyway .. I fucked that up .. damn

calispac: Is it solved ?

dneise:
Superseded by #116 and #117

baseline_stability.py
dneise: yes it is solved, but I closed this PR. Since it way too long. I replaced it with two other PRs, which are hopefully smaller and easier to review.

dneise: It was solved directly after you commented on it in this commit:
https://github.com/cta-sst-1m/digicampipe/pull/114/commits/f470706bce30ff7b1ced95b84a1380727e1e3dd1

PR #116, opened 2018-02-19
Nice event source and pep8 only

dneise:
(Should be) improved version of #114.  PR #114 was way way way too long to review. 
So I tried to tidy this up and do only what the name suggests.


dneise:
This is certainly not perfect ... but it is a step forward, I think.

baseline_stability.py
```diff
@@ -1,68 +1,32 @@
...
+from docopt import docopt
+from tqdm import tqdm
 
-    # Define the event stream
-    data_stream = event_stream(file_list=file_list, expert_mode=True, camera_geometry=digicam_geometry, camera=digicam)
 
-    for i, data in zip(range(n_events), data_stream):
+def main(infile):
 
+    baselines = []
```
calispac: I would prefer to initiate it as a numpy array as it was before: 
` digicam_baseline = np.zeros((len(digicam_geometry.pix_id), n_events)) `

then we can work on mean, std max and so on...
P.S. don't forget to change the append() in the for loop.

calispac: we can merge if you don't what to change it here

dneise: This solution does not need the number of events.

dneise: I make it an np.array after the loop okay? That way we still do not need the number of array, but can still use all the features of numpy

calispac: Ok for me!

PR #117, opened 2018-02-19
WIP: Split analysis and output

dneise:
**Work in progress**
So I worked on this quite a lot on the weekend ... and it resulted in a huge mess, I think.... 
This PR grew in size ... and it contains multiple different things. I would close it... but maybe somebody is interested ... 

This PR is about the distinction between "analysis", "serialization" and "graphical representation".

In order to write unit tests for an analysis it is better when the analysis result (or results) are not stored to disk, but just returned.

Storing something is a side-effect, and such side-effects are difficult to test. 

Some things are just difficult to test and at the same time, they are readily available from external libraries. Storing an array in a fits-file, json-file, hdf5-file, is nothing we need to worry about. So we also do not need to test it.

But what if the analysis throws an exception, because some input is not in an expected format or so? And our users come an complain: "why does this and that not work??". That is something we want to avoid. So we want to test our analysis code, not the code that is writing the results to disk.

The same is true for plotting. It is hard to test, if a plot is showing the data nicely on some users screen. So we let `matplotlib` do the job for us and do not test it.

However, when the only "output" of an analysis is the plot on the screen, then we cannot test this analysis. But when the analysis consists of two parts:
 * create the results and
 * plot it

in code similar to this:
```python
def main(a_lot_of_input_for_analysis, input_for_plotting):
    results = analysis(a_lot_of_input_for_analysis)
    plot_results(input_for_plotting)
```

Then for the tests, we can only test the `analysis()` function. And we are happy.


dneise:
This grew too large .. I will take interesting parts out and close this one.

PR #119, opened 2018-02-21
Processor Proposal

dneise:
Hello! :-D

So this PR is a proposal for a big interface change. @calispac asked me to present it, so here it goes...

These are the main points:
 * Analysis tasks are split into `Processor`s. 
 * complete analysis is just a group (`list`) of `Processor` instances.

A `Processor` is just a very thin base class, which just ensures, that every processor implements the `__call__` method, not more... 
https://github.com/cta-sst-1m/digicampipe/blob/a9a4a3b4b23f801494a39d45665740fc28d872c3/digicampipe/processors/__init__.py#L10-L20

So this means this function `func`:
```python
def func(event):
    return event
```
is also a valid processor. Even though it is not a class and it does not inherit from `Processor`, it fulfills the interface and that is important. This shows again the feature of the Python language called "duck typing" ... as long as is quacks like a duck and tastes like a duck, it probably is a duck. Same is true for `Processor`. 

We think of a processor like this: 
  * Take an `event`
  * to something with it
  * and write the result back into the `event`
  * (you may have internal state)

So a simple example processor might be this:
```python
class MyProcessor(digicampipe.processors.Processor):
  def __init__(self, some_parameter = 0.5):
      self.n_events_already_seen = 0
      self.some_parameter = some_parameter
  def __call__(self, event):
      self.n_events_already_seen += 1
      some_analysis_result = event.adc_samples.sum()
      some_analysis_result *= self.n_events_already_seen 
      some_analysis_result /= self.some_parameter
      event.sub_container.field_for_analysis_result =  some_analysis_result
      return Event
```

So you see:
 * Every processor has to find its input in the event. 
 * It can have some parameters, set by the user on instantiation. 
 * And it can have internal state, which changes while iterating over the events and has some impact on the result... if needed.
 * And it must write its output also into the event.

----

In [processors/pipeline_crab.py](https://github.com/cta-sst-1m/digicampipe/blob/processor/digicampipe/processors/pipeline_crab.py) you can find an example for an analysis.

You see the entire analysis is described using a list of processors. Nothing happens while defining this list of Processors. After the process is defined, one iterates over the events and applies each processor in the list process to the event in the same order as it was defined. 

I presume, in case we have a rather complete list of processors, users might plug and play with different analyses by either modifying processors parameters or by using different processors, but this loop over the events is not interesting. So I put this loop into a convenience function called `run_process(process, files)` I only mentioned in a comment on top of the loop.

------

### custom `SkipEvent` exception.

You know, when a Python iterable wants to signalize that it cannot iterate further, it simply raises the `StopIteration` exception. The for-loop catches this exception and ends. 

In our process we sometimes want to skip entire events, when we realize it makes no sense to look at this event at all. For this I've decided to implement the `SkipEvent` exception. When ever a processor raises this exception the process will continue with the next event.


dneise:
Ah .. nothing is tested here yet. 

In order to test this, I would in principle have to implement unit tests for the processors. The processors are derived from different functions, which all are not unit tested. So strictly speaking, I have converted these functions into processors, but I have no idea if they still work. Basically, because I have no idea if the original functions work. 

This is a big problem. We cannot argue about different code designs and play around with it, when we cannot be sure that the stuff still works. And it is impossible to define what "it works" means without tests. I've seen functions here, which I would define as broken even before I converted them into a Processor.

calispac:
Great thanks! I think it is the direction we want to take...
Just a few questions 

1. How can we deal with provenance of data? Should we have something that is a sub class of processor: 
```python
class DataSaver(Processor):

    def __init__(file):

        self.file = file
    
    def __call__(event, processors):
         file.write(event)
         file.add_provenance(processors)
         return event
``` 
2. Is there a way to make a list of processors a processor? Meaning that a chain of processes is a process

dneise:
The Provenance I wanted to handle in the Processor Baseclass ...  but I have not yet thought about it great detail.

dneise:
For grouping of processors into one processor I have several ideas: These solutions do not necessarily exclude each other.

One possibility would be to accept "sub-lists" simply this `processors.run_loop()` function.
```python
def run_process(process, files):
    for event in event_stream(files):
        try:
            process_event(process, event)
        except SkipEvent:
                continue


def process_event(processor, event):
    if isinstance(processor, Processor):
        event = processor(event)
    else:
        # if this is not a processor, it might be a list 
        # full of processors:
        for sub_processor in processor:
            event = process_event(sub_processor, event)
    return event
```

So when a user who has such a `process` at the moment:
```python
# file: my analysis.py
process = [
ProcA(),
ProcB(parameter=5),
ProcC(method='mean'),
SpecialProc(),
]
run_process(process, some_files)
```

realises, that the first 3 steps are always the same, she might decide to create a new file e.g. called `std_analysis.py` which looks like this:
```python
# std_analysis.py
std_analysis_start = [
ProcA(),
ProcB(parameter=5),
ProcC(method='mean'),
]
```

And then modify her current file like this:
```python
# file: my analysis_new_version.py
from std_analysis import std_analysis_start

process = [std_analysis_start, SpecialProc()]
run_process(process, some_files)
```

dneise:
Another Option for would be to not allow sub-lists, but create special processor, which can be initialized with a list of other processors, e.g. like this:

```python
std_analysis_start = processors.ProcessorGroup(list_of_processors=[
    ProcA(),
    ProcB(parameter=5),
    ProcC(method='mean'),
])
```

And this new class `ProcessorGroup` might look like this:
```python
class ProcessorGroup(Processor):

  def __init__(self, list_of_processors):
      self.list_of_processors = list_of_processors

  def __call__(self, event):
      for p in self.list_of_processors:
          event = p(event)
```

But ... now that I wrote it down, I like the first solution better ... since this new class `ProcessorGroup` is just a list in a Processor costume. 
It is also a new class ... and we should [Stop writing classes](https://www.youtube.com/watch?v=o9pEzgHorH0) :-D ...

dneise:
Ah I have an important question .. maybe to @calispac but maybe also to others.

I mean .. I like these processors (obviously) .. but I also like the generators which are currently in use mostly.... I think the difference is mostly of cosmetic nature ... 

So .. the question is: 

Should we go forward with this this? or is it **only cosmetic** and thus a waste of time?

dneise:
I would like to say something about the functions, which also fulfill the processor interface.

Suppose you have this analysis function:
```python

def integrate_adc_samples(event, method, integration_window=(10, 20)):
    # get input from event
    adc = event.r0.adc_samples

    # perform analysis 
    selection = slice(*integration_window)
    integral = method(adc[:, selection])
     
    # store result back into event.
    event.dl500.image = integral
    return event
```

How to convert this function into a Processor most easily?

calispac:
> Should we go forward with this this? or is it only cosmetic and thus a waste of time?

If the processor only process then I agree it is completely equivalent to a simpler function. But if for the sake of provenance we need some Processors then I think it is worth it. We might need to brainstorm on what a processor should be!

dneise:
The answer is using `functools.partial`

```python
from functools import partial

integration_processor_sum_big_window = partial(
    integrate_adc_samples,
    method=np.sum,
    integration_window=(0, 20)
)

integration_processor_trapz_smaller_window = partial(
    integrate_adc_samples,
    method=np.trapz,
    integration_window=(10, 20)
)

```

Both of these are now valid processors ... no need to implement any class. I think this is nice.

However .. the `event` is not so nice .. it wants us to always iterate over these  `tels_with_data` BS ... so all processors (if function or class does not matter) are much larger than they need to be.

dneise:
>  But if for the sake of provenance ...

Yes if provenance is important, then we gain from Processors, which are classes.

I think we should discuss, what we can expect from provenance and what not... I think ... this PR (which is called processor and which implements at the moment only the processor ... is not the right place to discuss provenance ... let's open an issue and discuss it there.

calispac:
> However .. the event is not so nice .. it wants us to always iterate over these tels_with_data BS ... so all processors (if function or class does not matter) are much larger than they need to be.

I also want to get rid of the `tels_with_data` thing... It a mess.

yrenier:
I probably missed something, but in the current state, what is the improvement beside not having to pass the event_stream around in the analysis code ?

dneise:
I guess we do not need this at the moment.

PR #121, opened 2018-02-21
Jurysek simtel2

jurysek:
This PR is digicampipe modification for processing sim_telarray raw output up to Hillas parameter to compare simulated data with real data processed in the same way.

The main code is **simtel_pipeline.py**, with the same structure like crab_pipeline.py from master. There are some new functions needed for saving additional informations about each event:

**events_image.save_events()** - it creates a TXT file with charge in each pixel that survived cleaning, one event per line. It is for machine learning purposes. It needs also a map of pixels, which is saved too by the same function.

**mc_shower.save_shower()** - it creates a TXT file with basic MC parameters about each processed event, that is needed for further analysis.

**events_image.save_timing()** - creates a TXT file with similar structure like .save_events(), but here is saved 'arrival time' (dl1_camera.time_bin[1]) in each pixel. It is needed for further timing analysis in some codes in jurysek_simtel branch, too be merged with master soon.

**simtel_baseline.save_mean_event_baseline()** - creates a TXT file with some statistic (mean of baselines for each event pixels that survived cleaning, and mean of std of baseline distribution) of calculated baseline. It is not needed for routine processing, thus commented.

There is also **simtel_baseline.fill_baseline_r0()** code for baseline computation directly from each waveform. The method is described as a comment directly in the code. 

The last, commented part of the code, is just for quick testing whether all pixels are mapped correctly and cleaned images looks somehow reasonable. I think it is good to let it there to check if sim_telarray camera_config_file is still compatible with digicampipe camera geometry from time to time. But we can also create some specific function for these purposes..

For reading of the sim_telarray output files in eventio format, Python package pyhessio with some small modifications (**digicampipe/io/hessio_digicam.py**) is used in order to fill the containers so that the event stream from simulated data is very similar to event stream from real data zfits files. Some temporary solutions like line 109 (allowed_tels = None) and so on are terrible, I know..

Data containers **containers_mc.py** are taken from ctapipe containers.py include also containers for MC data, just slightly modified for digicam purposes. The best way would be just to add those MC containers to digicam containers, but I wanted to avoid touching master as much as possible.


dneise:
Thanks for opening this PR. Are you ready for some feeedback?

Anyway the first feedback I can already give is: Please have a look if you can write your code according to PEP8.

Depending on the editor you are using, there are plugins, showing you PEP8 issues already while typing ... these exist for vim and emacs as well as for PyCharm.

dneise:
Ah I was too late you are already starting ... perfect thanks a lot.

dneise:
> event_stream.py - more than 2 choices needed

Yes I agree ... ctapipe has solved this with a factory. Which works, but looks very code-heavy. I have something in mind .. I make a proposal ... gimme a sec.

dneise:
> utils.integrate - original return h[...,int(np.floor((window_width-1)/2)):-int(np.floor(window_width/2))]
is not equivalent to return h[..., (window_width-1)//2:-window_width//2]

Oh it is not? you are right ..

jurysek:
No, but we can of course change it just like:

return h[..., (window_width-1)//2:-(window_width//2)]

dneise:
> No, but we can of course change it just like: return h[..., (window_width-1)//2:-(window_width//2)]

Well we can do whatever we want, but I must admit at the moment, I do not know what this code is **supposed to do**... I thought it was easy, but it is apparently not .. since I fucked it up.

dneise:
Ah now I get your point, it is about the brackets ... integer division is different for negative numbers ... 
You are right, the smallest correct way is indeed this:
```
 (window_width-1)//2:-(window_width//2)
```

jurysek:
If I remember it well from my thorough reading of the code in the past, 
I think it is just an integration via convolution with unit step 
function, which gives you not only the integral itself, but also a 
position of maximum. But I don't remember why is needed to remove a half 
of the window from both sides of 'h' after it..

dneise:
Ah the borders need to be removed, because the convolution kernel does not know what to do at the borders.

jurysek:
Oh, sure ... OK, so should I push the branch again with repaired brackets?

dneise:
yes please sounds perfect

dneise:
I have to deal with some other stuff now ... so probably no more review from me for today.

jurysek:
Sure, thanks a lot for the review! There just 2 things I touched from 
master left, so we can solve it tomorrow.

digicampipe/calib/camera/dl1.py
```diff
@@ -72,7 +73,21 @@ def calibrate_to_dl1(event_stream, time_integration_options, picture_threshold=7
 
             dl1_camera.on_border = border
 
-            dl1_camera.cleaning_mask = cleaning.dilate(geom=geom, mask=dl1_camera.cleaning_mask)
+            """
+            # repaired piece of code from Cyril. The commented version above leads to border_flag = 1 in almost all events.
```
dneise: I must admit I have absolutely no idea about this part of the code. I guess @calispac should comment.

I believe you, when you say the previous version in the master was simply broken and this works now. I would be great  if you could add a test, which shows this is working now (and if the test would be thrown against the old version, it would show that it does not work).

But if that is too much work for now ... having working code is always better than having broken code. Thanks for fixing it.

digicampipe/io/save_hillas.py
```diff
@@ -34,14 +35,15 @@ def save_hillas_parameters(data_stream, n_showers, output_filename):
...
+        border.append(1 if event.dl1.tel[1].on_border else 0)
 
         print('hillas #', i)
 
     np.savez(output_filename, size=size, cen_x=cen_x, cen_y=cen_y, length=length, width=width, r=r, phi=phi, psi=psi,
              miss=miss, skewness=skewness, kurtosis=kurtosis, event_number=event_number, time_stamp=time_stamp,
-             time_spread=time_spread)
-    
-    
+             time_spread=time_spread, border=border)
```
dneise: very okay for me. @calispac ? Or for you as well?

dneise:
Okay so.. there are still a couple of pep8 problems ... 
I will check this branch out now an read the code in more detail ... in the process, I will fix the pep8 problems ... should I push that in here directly? or should I make another mini pull request?

jurysek:
Thanks a lot! I didn't want to interfere too much with the code in master, thus I solved some pep8 issues just in the code I added. You can push Your changes in here directly, I do not plan to edit anything before merge this branch with master.

dneise:
okay I did a mini PR anyway .. you can review it or not ... currently tests are failing due to an unconnected problem with a package, which might be removed from the anaconda server or so .. I am looking into that.

dneise:
Ah another thing... after the conflicts with the master have been discussed, it is somehow our task to review the actual feature you are planning to introduce in this PR.

By the sheer number of lines you have added/modified (~1.3k) it might be understandable, that reviewing this PR might take quite some time. 

So could you maybe summarize at the top. What the new feature is .. or the new features are, that you are implementing here? That might be helpful, to guide us.

jurysek:
Sure, I am going to describe it in the comment window above.

dneise:
Awesome thanks

simtel_baseline.py
```diff
@@ -0,0 +1,62 @@
+import numpy as np
+
+
+def fill_baseline_r0(event_stream, n_bins0=5, n_bins1=10, method='data'):
+
+    for count, event in enumerate(event_stream):
```
dneise: `count` is unused .. could be removed.

simtel_baseline.py
```diff
@@ -0,0 +1,62 @@
...
+
+        for telescope_id in event.r0.tels_with_data:
+
+            if method == 'data':
```
dneise: this "if method .. else other_method" could also be implemented as two independend functions.

dneise:
In some places there are still comments, like these:
```python
# import inspect
# import pyhessio
# print(inspect.getfile(pyhessio))
```

These comments look like somebody is still in the middle of playing around with the code ... testing this .. testing that. I creates an untidy impression. This is totally fine for work in progress, but it should stay out of the master if possible.

jurysek:
You are right, I didn't remove all of these testing parts.. Should I do it and push, or You are just in the middle of the process?

dneise:
I am just reading .. you can push .. I'll every now and then. .

simtel_pipeline.py
```diff
@@ -0,0 +1,279 @@
...
+    # Save Hillas
+    hillas_filename = options.output + '_' + suffix
+    save_hillas_parameters(
+        data_stream=data_stream,
+        n_showers=n_showers,
+        output_filename=directory + hillas_filename)
+    # save_hillas_parameters_in_text(
+    #    data_stream=data_stream,
+    #    output_filename=directory + hillas_filename)
+
```
dneise: This is the main script of this PR. In the bottom is a lot of commented out plotting code. I find it great to split plotting from doing the analysis. So .. can this be removed?

pipeline_crab.py has evolved a bit, since this was written, I assume. So they look very different now. 
I assume the goal should be to have one "pipeline.py", which can eat zfits or hessio and knows how to treat any of them, right? But I assume that willl be done in a future step, correct? 

In that case I think this is fine here.

simtel_baseline.py
```diff
@@ -0,0 +1,62 @@
+import numpy as np
+
+
+def fill_baseline_r0(event_stream, n_bins0=5, n_bins1=10, method='data'):
```
dneise: Ah .. I just see that this function is basically modelled after `digicampipe.calib.camera.random_triggers.fill_baseline_r0()`, right?

I assume you had to write your own function here, since the other `fill_baseline_r0` did somehow not work for you. I think this is fine, but we should merge these two functions as soon as possible ... otherwise we end up with 3 functions for every task, one for zfits one for hdf5 one for hession ... and each of these functions might work **nearly** identical .. but not quite .. that quickly turns into chaos.... I fear.

We assume we can do this later .. but we should keep it in mind. If we merge this as it is, I'd be happy if we could open an issue, kind of as a reminder to merge this stuff.

mc_shower.py
```diff
@@ -0,0 +1,35 @@
...
+
+        mc_data = np.hstack((event_id, run_id, core_distance, energy,
+                            theta, phi, h_first_int, offset_fov_x, offset_fov_y,
+                            core_x, core_y))
+        mc_data_all.append(mc_data)
+
+        yield event
+
+    # save mc parameters of showers for all events
+    save_file(mc_data_all, filename_showerparam)
```
dneise: I think this is new ... so you can do whatever you like here.

For completely new features a unit test would be nice ..

dneise:
@jurysek I approved this just now, but still had a couple of comments. 
I'd still be happy to have @calispac take a look as well, since this PR touches some code, I was not totally familiar with.

digicampipe/calib/camera/dl1.py
calispac: its fine although we should move this piece of code to a function I think

digicampipe/io/event_stream.py
```diff
@@ -33,6 +31,15 @@ def event_stream(
...
 
+def guess_source_from_path(path):
+    if path.endswith('.fits.fz'):
+        return zfits.zfits_event_source
+    elif path.endswith('.h5') or path.endswith('.hdf5'):
+        return hdf5.digicamtoy_event_source
+    else:
+        return hsm.hessio_event_source
+
+
```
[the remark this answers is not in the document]
calispac: Just to comment on this I realized that now all `*event_source` stuff need to take the same arguments otherwise it breaks.

For instance we remove the `camera` argument in `zfits_event_source` the we also need to remove it from this function and from the others (digicamtoy, hessio). What if we just do: 

```python
event_stream(filelist, **kwargs)
```
then pass `**kwargs` to the `event_source` stuffs?

digicampipe/io/event_stream.py
```diff
@@ -9,20 +10,17 @@ def event_stream(
     camera_geometry=None,
     expert_mode=None,
     max_events=None,
-    mc=False
+    source=None
```
calispac: if source is not None then this function does not do anything I believe. we can remove source as arg.

Then latter we try to make guess_souce_from_path() smarter

digicampipe/io/hessio_digicam.py
```diff
@@ -0,0 +1,297 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+"""
+Components to read HESSIO data.
+
+This requires the hessio python library to be installed
```
calispac: Is this part of the requirements now? or does it install with ctapipe?

calispac: Also seems like the entire file is from ctapipe. What can be imported and not duplicated?

digicampipe/io/event_stream.py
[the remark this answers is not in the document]
dneise: sounds good. I'm doing it

digicampipe/io/save_hillas.py
```diff
@@ -34,14 +35,15 @@ def save_hillas_parameters(data_stream, n_showers, output_filename):
         event_number.append(event.r0.event_id)
         time_spread.append(event.dl1.tel[1].time_spread)
         time_stamp.append(event.r0.tel[1].local_camera_clock)
+        border.append(1 if event.dl1.tel[1].on_border else 0)
```
calispac: maybe tel[1] is not the best here... 
I would suggest to just ask for the telescope_id for this function argument

Also why don't save directly event.dl1.tel[1].on_border ?

the `1 if event.dl1.tel[1].on_border else 0` seems confusing.

mc_shower.py
calispac: At some point should we come to a save_everything() function ?

simtel_baseline.py
calispac: yes lets_make to functions and the the main one does it the `sim_tel` way and the `data` way

simtel_pipeline.py
```diff
@@ -0,0 +1,292 @@
...
+    parser.add_option('-d', "--baseline0", dest='baseline_beginning',
+                      help='N bins from the beginning of the waveform for \
+                      baseline calculation', type=int, default=9)
+    parser.add_option('-e', "--baseline1", dest='baseline_end',
+                      help='N bins from the end of the waveform for baseline \
+                      calculation', type=int, default=15)
+
+    (options, args) = parser.parse_args()
+
+    digicam_config_file = 'camera_config_digicampipe.cfg'
```
calispac: this is overwritten at line 72 so no need for it here?

simtel_pipeline.py
```diff
@@ -0,0 +1,292 @@
...
+
+    # Save cleaned events - pixels and corresponding values
+    filename_pix = 'pixels.txt'
+    filename_eventsimage = 'events_image_' + suffix + '.txt'
+    data_stream = events_image.save_events(
+        data_stream, directory + filename_pix,
+        directory + filename_eventsimage)
+
+    # Save simulated shower paramters
+    filename_showerparam = 'pipedmc_param_' + suffix + '.txt'
```
calispac: At some point we should create an output format for the analysis.

digicampipe/io/hessio_digicam.py
```diff
@@ -0,0 +1,297 @@
...
+
+    Returns
+    -------
+    event_id_list : list[num_events]
+        A list with all the event ids that are in the file.
+
+    """
+    logger.warning("This method is slow. Need to find faster method.")
+    try:
+        with open_hessio(url) as pyhessio:
```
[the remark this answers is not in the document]
dneise: Actually when the context "hessio" is already clear .. since we are in a file called "hessio_digicam" .. instead of "hessio_file" ... `file` would be good as well.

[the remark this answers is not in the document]
jurysek: I agree. I changed it and I will push the repaired code in a moment.

digicampipe/io/hessio_digicam.py
```diff
@@ -0,0 +1,297 @@
...
+            # tels_with_data and allowed_tels
+            if allowed_tels is not None:
+                selected = data.r0.tels_with_data & allowed_tels
+                if len(selected) == 0:
+                    continue  # skip event
+                data.r0.tels_with_data = selected
+                data.r1.tels_with_data = selected
+                data.dl0.tels_with_data = selected
+
+            data.trig.tels_with_trigger \
```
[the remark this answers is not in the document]
jurysek: Well, most of the code is copy/paste from ctapipe.io.hessio, so I don't know if is reasonable to change it with respect to future merge with ctapipe..

digicampipe/io/containers_mc.py
```diff
@@ -0,0 +1,518 @@
+"""
+Container structures for data that should be read or written to disk
+"""
```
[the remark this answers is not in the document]
jurysek: I wasn't sure, because io.containers_mc are in fact containers from ctapipe with addition of some containers specific for digicampipe. I don't know what is actual plan about merge digicampipe with ctapipe and, therefore, which way is better. That is why I prepared containers_mc which should be (I hope) compatible with both.

simtel_pipeline.py
jurysek: Well, yes, the commented code is, as described at the top, just for testing compatibility between simtel and digicampipe pixel map, because I had some problems with it before. It is definitely just a temporary code and it will be surely removed but I would plead for keeping it for a while..

Oh, I see.. Sure, the final goal is to have one universal pipeline. But at this point, when we are still testing some settings specific for simulated data like baseline determination, I would say that separated pipe for simulations, which uses general functions and classes from digicampipe is less painful way.

simtel_pipeline.py
jurysek: Oh, sure, corrected..

simtel_pipeline.py
jurysek: I completely agree with You. The final format of those files should be discussed and probably something better than a lot of TXT output files should be proposed..

digicampipe/io/containers_mc.py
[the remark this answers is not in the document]
dneise: I must admit .. I am totally lost in this discussion ... containers in ctapipe will be modified a lot ... so whatever we have here now .. will not help us much when merging I guess.
Let's just do what works...

simtel_pipeline.py
dneise: >  ... just for testing compatibility ...

This explanation virtually screams "make me a unit test... please!". :-D

simtel_pipeline.py
dneise: No need to define something *final*, I think. This sounds like, we have to use txt until we know **for sure** what will be final. But we will never know anything for sure ... Still we can make a little improvement any time we want to make one .. event when it is not final.

simtel_baseline.py
jurysek: Agree.. for now I I have just renamed and split the two methods into two functions

digicampipe/io/containers_mc.py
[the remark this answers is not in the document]
jurysek: OK, so in that case I have made digicampipe.io.containers compatible with simtel_pipeline and we don't need containers_mc anymore. Please check if I haven't destroyed anything..

digicampipe/io/hessio_digicam.py
jurysek: > Is this part of the requirements now? or does it install with ctapipe?
Everything what it needs is installed with ctapipe.

> Also seems like the entire file is from ctapipe. What can be imported and not duplicated?
Correct, but the problem is that digicampipe needs some more containers to be filled than original ctapipe.io.hessio supports and since it is written like several long functions, I just took the functions and added a few more lines inside each of them.

digicampipe/io/save_hillas.py
calispac: ok

digicampipe/io/containers_mc.py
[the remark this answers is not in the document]
dneise: I don't know how to check that, we still lack test for this I believe.

digicampipe/io/containers.py
```diff
@@ -180,19 +297,22 @@ class ReconstructedShowerContainer(Container):
...
+    az_uncertainty = Field(0.0, 'reconstructed azimuth uncertainty',
+                           unit=u.deg)
     core_x = Field(0.0, 'reconstructed x coordinate of the core position',
                   unit=u.m)
     core_y = Field(0.0, 'reconstructed y coordinate of the core position',
                   unit=u.m)
-    core_uncertainty = Field(0.0, 'uncertainty of the reconstructed core position',
-                       unit=u.m)
+    core_uncertainty = Field(0.0, 'uncertainty of the reconstructed core \
+                             position', unit=u.m)
```
dneise: please avoid \-backslash line continuation...

dneise:
Jakub asked me in a mail, if this can be merged now ... He has more stuff to put into the master and in order to proceed this needs to be merged...

I think it is fine ... @calispac might want to comment.

I just had a quick look at the containers and it looks like there were only additions .. nothing which existed before was removed, so everything else should still work.

I'd say we can merge this.

calispac:
> I'd say we can merge this.

Yes let's go ahead

PR #123, opened 2018-02-21
guess event_source from path unless given explicitely

dneise:
This is a small change proposed for branch `jurysek_simtel2` in PR #121 

The event_stream simply guesses which source to use from the given path. 
This is a **very simplistic** implementation to guess which source to use, but as long is it works, it is fine I guess.

Another implementation would maybe check the first few bytes, as mentioned here:
https://en.wikipedia.org/wiki/List_of_file_signatures

This does not contain the magic number for hdf5, but it is mentioned here:
https://en.wikipedia.org/wiki/Hierarchical_Data_Format

for Hessio, I have no idea which magic numbers to use.

So .. anyway as long as guess the type from the path works, we might run with a solution like this. ... 
What do you think?

### DIFF
--- a/digicampipe/io/event_stream.py
+++ b/digicampipe/io/event_stream.py
@@ -10,7 +10,7 @@ def event_stream(
     camera_geometry=None,
     expert_mode=None,
     max_events=None,
-    mc='real'
+    source=None
 ):
     # If the caller gives us a path and not a list of paths,
     # we convert it to a list.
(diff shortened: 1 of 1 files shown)


moderski:
It is quite dangerous to assume that all sim_tel output has an extension of .simtel.gz, especially that no such file is generated by default (you need to specify OUTPUT_FILE option).
BTW the eventio file format used by sim_telarray is described in attached document (from the hessio package distributed with sim_telarray).
[eventio_en.pdf](https://github.com/cta-sst-1m/digicampipe/files/1746974/eventio_en.pdf)

PR #126, opened 2018-02-23
Jurysek simtel2 pep8 dneise

dneise:
only pep8


dneise:
Ok .. pep8 is good now .. I did not shorten the lines around the "border_flag" stuff ... this comment will be removed anyway soon, so no need to shorten it.

Issue #129, opened 2018-02-23
error during documentation buil


moderski:
When you hit the "build" button in the README.md section of the https://github.com/cta-sst-1m/digicampipe page, after successful build of the package itself, there is a complain on documentation build on Travis:
```
$ make -C docs/ html
make: Entering directory `/home/travis/build/cta-sst-1m/digicampipe/docs'
Running Sphinx v1.6.2
making output directory...
Extension error:
Could not import extension sphinx_automodapi.automodapi (exception: No module named 'sphinx_automodapi')
make: *** [html] Error 1
make: Leaving directory `/home/travis/build/cta-sst-1m/digicampipe/docs'
The command "make -C docs/ html" exited with 2.
```

dneise:
`sphinx_automodapi` was installable from conda, but it not anymore ... 
needs to be pip installed now.

So our master broke without us touching it, because a dependency was modified.

I ran into this problem in this PR #126 and am trying to fix it there ... 
#126 will be merged into #121 ... which is still under review .. and since it is a large PR, it might take ... dunno .. a week? until it is merged.

So we can either wait until #121 is merged .. or make a little fix only for this...

dneise:
I'll make a small fix for this ...

PR #130, opened 2018-02-23
Install sphinx_automodapi using pip.

dneise:
Also updating from v0.5 to v0.7 ... 

Fix for #129 

Waiting for tests  to pass, I guess we can merge then


dneise:
I merge this without a review ... it is a really small hotfix for #129

PR #131, opened 2018-02-23
avoid keyword argument 'file_list'

sagreus:
Just edited some scripts to avoid the keyword argument 'file_list' in event_stream


baseline.py
```diff
@@ -25,7 +25,7 @@
 
 def main(baseline_file_path, files, unwanted_pixels=[]):
 
-    data_stream = event_stream(file_list=files)
+    data_stream = event_stream(files)
```
calispac: I think here we should just keep the name `file_list` not `files`

dneise: I just realize both a wrong.

event_stream happily eats any sequence of *paths* (not files), this includes
 * lists,
 * tuples
 * any iterable, which spits out paths, like e.g. `glob.iglob("*.fits.fz")`

So while I think `files` beats `file_list` here ... `paths` would be best.

----

We use `file_list` many of the other scripts ... but I think `paths` would be best.

sagreus: Ok, but then we should edit it also in:
`files=args['<files>'],` 
and
`def main(baseline_file_path, files, unwanted_pixels=[]): `

dneise:
@sagreus thanks for this PR!

There are many pep8 issues in this PR but IMO you do not have to solve them. Can be solved later.

So once we agreed on the question `file_list` vs. `files` vs `paths` ... would you have time to adjust the scripts, so they look consistent?

sagreus:
Sure I can do that, just let me know what you agree. 
I can also solve the pep8 issues too, they all look like trivial fixes.

dneise:
> I can also solve the pep8 issues too, they all look like trivial fixes.

sure ... go ahead .. pep8-ing is always welcome.

baseline.py
dneise: yep.

just in the help, we should stick to "<files>" I think, since it seems to be common in linux man pages to talk about "FILES" instead of "PATHS", c.f. [man page of rm](https://linux.die.net/man/1/rm).

So using "FILE" and "PATH" as as synonyms is fine on the command-line. But in source code there is a difference between a file and its path.

dneise: Ah another option would be `urls`

calispac: I think what is most important is the consistency of variable names

sagreus:
I cleaned all the pep8 warnings. Hope it looks ok. 
Waiting now for the decision on files, file_list, etc. To update again the pull request

dneise:
:-D ... damn politeness  ... nobody wants to take a decision :-D

So Cyril thinks consistency is very important, I agree, so whatever we do .. (you do :-D) .. is should be one everywhere.

I think `paths` or `urls` is best ..

dneise:
okay, so instead of telling @sagreus what to do, I did it myself, I guess that was ok.

So now the iterables full of file paths, are consistently named `urls`

sagreus:
Sorry I got distracted with other stuff. Sure, it is ok. Thanks.

dneise:
Can we merge this now?

PR #136, opened 2018-02-26
move viewer into pipe

### LINKED ISSUE #133: Should we move the viewer into digicampipe?
digicampipe and the viewer are not really disconnected:
 * Neither the viewer is independent from the pipe,
 * nor the pipe is independent from the viewer. 

c.f. https://github.com/cta-sst-1m/digicamviewer/issues/5

So putting them into 2 packages is not really helpful I think.

dneise:
fixes #133


dneise:
Ah .. wait don't merge yet

dneise:
okay I like it like this:
```python
viewer = EventViewer(data_stream)
viewer.draw()
```

nothing else needed to look at events ...

dneise:
Actually .. this is still one line too much .. could as well be: `view(data_stream)`

dneise:
But that we can do in a future PR.

dneise:
okay so this PR does not only move the viewer into digicampipe ..
it also creates a quick example.

After installation of this version with `pip install` or `pip install -e` a so called console_script is installed in your console...

so you can in any folder, where ever you are do:
```
digicamview SST1M01_20171030.022.fits.fz
```

And look at the events ..

dneise:
This is: 
 * a quick test if the viewer works and
 * a quick way to introduce newbies into digicam data.

dneise:
Okay I like this now .. 

have a look at the README ... 

https://github.com/cta-sst-1m/digicampipe/tree/eventviewer_into_pipe

PR #141, opened 2018-02-26
jurysek simtel2 - separation and reconstruction

jurysek:
This PR contains several codes for analysis, gamma/hadron separation and shower direction reconstruction for observations in single telescope mode.

# Code description

**rswl-generate-lookup.py** - generates lookup tables based on sim_telarray simulations for common gamma/hadron separation method - cuts on RSW and RSL. The tables are binned in impact parameter and log10(size). They should be generated for each zenith and azimuth separately. Some bins in the tables are not populated with any events with respect to an energy and impact distribution of triggered events.

**energy-generate-lookup.py** - generate lookup tables for energy reconstruction

**shower_geometry.py** - contains functions for calculation of impact parameter

**rswl-cuts-optimization.py** - Calculates RSW, RSL, efficiency and quality factors for gammas and protons with the use of the lookup tables and optimizes cuts for separation.

**rswl_plot.py** - plotting lookup tables, normalized distributions of RSW and RSL for gammas and protons, efficiency and quality factor dependency on RSW/L cuts.

**fill_lookup.py** - a function for binning simulated data in impact parameter and log10(size). It returns mean, std and N of data in each bin.


shower_geometry.py
```diff
@@ -0,0 +1,30 @@
+import numpy as np
+
+
+def line_point_distance(xp1, yp1, zp1, cx, cy, cz, x, y, z):
```
dneise: it might be worth to use something like 1D-np.arrays here, I guess. So that 
 * `p1` - point on line
 * `c` - direction of line
 * `p` - point in space

Then the code would reduce to:
```python
a = np.cross((p - p1), c)
return np.sqrt(np.sum(a**2) / np.sum(c**2))
```

rswl_plot.py
```diff
@@ -0,0 +1,132 @@
...
+import matplotlib.pyplot as plt
+import matplotlib as mpl
+
+
+def lookup2d(data):
+
+    data_width = data[:, [0, 1, 2]]
+    data_length = data[:, [0, 1, 3]]
+    data_sigmaw = data[:, [0, 1, 4]]
+    data_sigmal = data[:, [0, 1, 5]]
```
dneise: integer based element access from the 2D-array(?) called `data` here looks really cumbersome and error prone. It seems to be very easy to mess up something by using a wrong number, like:
```
data_sigmaw = data[:, [0, 2, 4]]
```

Also the reader of `lookup2d` has no idea what in columns [0, 1] might be... this could be maybe improved by using either numpy recarray or numpy structured array or pandas dataframe... or even more, but those are the ones I know about. pandas dataframe happens to plot itself quite nicely (sometimes) and it has an index column. I assume the columns 0 and 1 of `data` are some kind of index, which is why every `data_foo_bar` is looking at them

rswl_plot.py
```diff
@@ -0,0 +1,132 @@
...
+
+    width = data_width[:, 2].reshape(
+                                     (len(np.unique(data_width[:, 0])),
+                                      len(np.unique(data_width[:, 1]))
+                                      ))
+
```
dneise: oh no! .. no I understand...  `data[:, [0, 1]]` are shapes .. and they should all be identical .. ok this is probably still under construction here. forget my previous comment...

rswl-generate-lookup.py
```diff
@@ -0,0 +1,127 @@
...
+
+    return binned_wls
+
+
+if __name__ == '__main__':
+
+    parser = OptionParser()
+    parser.add_option('-g', '--hillas', dest='hillas_gamma',
+                      help='path to a file with hillas parameters of gamma',
+                      default='../../../sst-1m_simulace/data_test/ryzen_testprod/0.0deg/Data/hillas_gamma_ze00_az000_p13_b07.npz')
```
dneise: This will probably work only for one person in the world.

dneise: Oh .. this is a test, I assume.

shower_geometry.py
jurysek: OK, thanks, it looks way more better :-)

rswl_plot.py
jurysek: OK, I have changed the table to be structured array with named columns. It is much more transparent.

dneise:
Oh I should say: You don't have to follow all of my comments here, right? Sometimes I just emit "words" while reading your code .. Sorry I'll have to learn to keep my mouth shut.

jurysek:
No, it's completely OK as long as Your comments lead to an improvements of the code which is definitely the case now! So, don't worry and keep talking :-) It sometimes motivates me to look at the code closely again and repair some weird stuff :-)

dneise:
I will have no time to review this today ... while it is a long time open already .. maybe also tomorrow no time. Anybody else?

fill_lookup.py
```diff
@@ -0,0 +1,56 @@
+import numpy as np
+
+
+def fill_lookup(size_bins_edges, impact_bins_edges,
```
calispac: In principle this function can avoid the use of for loops using `axis=` feature and numpy array broadcasting.  But here I am not really sure how to help I would need to try it

rswl-cuts-optimization.py
```diff
@@ -0,0 +1,175 @@
...
+    rsl = rsl[~np.isnan(rsl)*~np.isinf(rsl)]
+
+    return rsw, rsl
+
+
+def efficiency_comp(gamma_cut, rswl):
+
+    efficiency = []
+
+    for ghc in gamma_cut:
```
calispac: does this work in replacement (avoiding for loops and taking advantage of numpy) ?

```python
bellow_gamma_cut = rswl < gamma_cut[..., np.newaxis]
efficiency = np.sum(bellow_gamma_cut, axis=0)
efficiency /= rswl.shape[0]
```

shower_geometry.py
```diff
@@ -0,0 +1,33 @@
...
+def impact_parameter(x_core, y_core, telpos, theta, phi):
+
+    cx = np.sin(np.deg2rad(theta))*np.cos(np.deg2rad(phi))  # direction cosines
+    cy = np.sin(np.deg2rad(theta))*np.sin(np.deg2rad(phi))  #
+    cz = np.cos(np.deg2rad(theta))                          #
+
+    direction = np.array([cx, cy, cz]).T
+    impact_point = np.array([x_core, y_core, np.zeros(x_core.shape[0])]).T
+
+    impact = []
```
calispac: Is this working? 

```python
impact = line_point_distance(impact_point, direction, telpos) 
``` 
If not maybe the axis should be specified in : 
```python
def line_point_distance(p1, c, p):
...
return np.sqrt(np.sum(a**2, axis=-1)/np.sum(c**2, axis=-1))
```

jurysek:
Thanks for the review! I have taken into account most of Cyril's suggestions in the new commit.

I planed to add some more codes into this PR, but since I have been working on something else for last couple of days, I think it would be better to merge this PR and continue in a separate one later.

calispac:
Fine for me since we can do profiling later.

Issue #146, opened 2018-02-28
digicam_baseline vs baseline_from_random_triggers


dneise:
After the interesting discussion about the baseline in context with the "R1 & DL0 Telescope Event Interfaces and Prototype Evaluation" document, I wanted to execute the request from @Etienne12345 s email which said:
```
TODO
    * Someone: estimate degradation of sensitivity induced by one baseline every 2 seconds rather than for every event
```

So .... I went ahead and wanted to see the impact on the hillas parameters in the crab data sample from last november. 
So I executed `pipeline_crab.py` on the data sample and then went on to modify the pipeline in such a way that the baseline is only updated once every 2 seconds. Only then I realized that the pipeline *is not using the digicam baseline at all*. 

Is this because `pipeline_crab.py` is not our "std analysis" but just some possible example?

If so, do we have a std-analysis? And if so ... why is it not implemented in pipeline_crab.py? Wouldn't that be a good place to document what we think at the moment is the best way to analyse our digicam data?

Etienne12345:
Indeed, afaik we use the baseline calculated from the interleaved triggers. I guess that we do so because the per-event baseline wasn't available since the very beginning, and nobody took the time to add it to digicampipe yet. I guess that if you feel like it you can go ahead and add it (tbc with Cyril of course).

calispac:
@dneise you can go ahead with what you propose. This is more evaluating the impact on the full reconstruction. 

I was thinking of doing it using Monte Carlo. We would evaluate the baseline precision as function of NSB rate and number of samples used to compute the baseline.

The "monitoring" baseline method is sort of equivalent to the "DigiCam" baseline method if the rate of events is sufficiently high enough.

dneise:
I had a look the different methods, in case you are interested:
https://github.com/dneise/digicam_crab_data_study/blob/master/Compare_digicam_and_pedestal_baseline.ipynb

dneise:
Okay so in this notebook I am starting to compare the hillas parameters for the different baseline strategies.

It takes quite a long time to run and I am not yet done. But in order to perform this study, I think I had to modify digicampipe as well. 

So together with this notebook comes a digigcampipe branch (and a PR) ... so despite this not being ready yet... here is the notebook and a link to the branch.

 * [notebook](https://github.com/dneise/digicam_crab_data_study/blob/master/Compare_hillas_parameters_using_different_baseline_strategies.ipynb)
[...]

PR #147, opened 2018-03-04
WIP: stuff I need for my crab analysis

dneise:
work in progress


dneise:
The analysis turned out to be not very helpful, c.f. @moderski s comment in #146, so there is no real reason to merge this. However, I wanted to experiment in this branch a bit with the question: 

How to modify parts of the analysis in a way, that I do not have to copy large parts of the code base.

As you see in the current situation, I just copied+pasted the parts where I needed to make a modification. 
I think from a functional point of view, this code is doing exactly what I wanted. 
So now I can basically invent a functional test, which checks that the output does not change, and modify the code until I can get the same result, without doing so much copy+paste. 

I think this is an interesting study for myself, and maybe for others as well.

PR #149, opened 2018-03-06
runtime  improvement

dneise:
`dark_baseline['baseline']` in this place is not a cheap attribute-get, but an expensive disk hitting action.

Doing this in a for-loop for every event is pretty expensive.

I did not know that in the first place. Learned it only by profiling ...


dneise:
Nothing urgent here just did not want to forget about it

dneise:
After this the time is distributed like this:
 * 45% zfits event source
    * 33% rawzfitsreader.readEvent, 
    * 6% _prepare_trigger_output, 
    * 2% _prepare_trigger_input, 
    * 4% ?
 * 23.5% calibrate_to_dl1
    * 13.5% ctapipe.cleaning.tailcuts_clean
    * 18% ["breakout_start_part"](https://github.com/cta-sst-1m/digicampipe/blob/7aea08ee7c7456bad7c7f91eee61183dca3ffe21/digicampipe/calib/camera/dl1.py#L119-L144)
  
No real idea where the rest of the time really goes :-D here and there...

dneise:
One thing, where we could save a bit of time is of course making these "trigger_output" and "trigger_input" stuff lazy, so that it is only calculated, when really needed. But that is not really possible with ctapipe.Containers, so going this path involves many lines of code for maybe ~8% speed up, which is imo not worth it.

In this "calibrate_to_dl1 start part", the integration costs 7%, np.pad costs ~4.5% and the rest (6.5%) is somwhere hidden in that start_part ... I am not sure what is so costly there... 

Ah maybe I should say what the current (single core) analysis rate is: 
So my CPU needs ~4.4ms for one event resulting in ~230Hz analysis rate.

So when the camera triggers with 10kHz we would at the moment need ~44 cores in the camera server to analyze in real time. To be honest I have no idea what the camera server is supposed to do, but I think it is somehow nice, if an analysis can be performed in realtime on the camera server as the events are coming in. I assume something in the order of 16 cores is normal nowadays and 32 cores is also well possible. So with 44 cores we are maybe ~40% too slow for my personal goal :-D. I think that is okay given that we did not optimize at all so far.

So I think we are not in a bad shape at all at the moment. Congrats everybody!

dneise:
Ah I just found why ctapipe.cleaning.tailcuts_clean is so slow ... 

I think we can gain another 13.5% speedup, without an adverse effect on code readability. I will open a PR in ctapipe for that.

I think we might want to think about updating our ctapipe dependency.

digicampipe/calib/camera/r1.py
```diff
@@ -3,7 +3,7 @@
 
 
 def calibrate_to_r1(event_stream, dark_baseline):
-
+    dark_baseline = dark_baseline['baseline']
```
calispac: Thanks for the improvements

calispac: I would rather prefer actually that we pass the `dark_baseline['baseline']` instead of the `dict()` /`.npz` file containing the baseline (`dark_baseline`). I think this solves the issue (i.e. loaded once outside the function) correct?

Also I think it is more understandable to pass an array rather than a dict.
What do you think?

digicampipe/calib/camera/r1.py
```diff
@@ -26,7 +26,7 @@ def calibrate_to_r1(event_stream, dark_baseline):
                 r1_camera.nsb = calib.compute_nsb_rate(standard_deviation, 'std')
             else:
                 # compute NSB and Gain drop from baseline shift
-                r0_camera.dark_baseline = dark_baseline['baseline']
+                r0_camera.dark_baseline = dark_baseline
```
calispac: This block is the else statement of : 

```python
if dark baseline is None:
```
with the change in the code now it is not doing the same. Previously it was checking if the argument passed in the func. `dark_baseline` was `None`. Now it checks if `dark_baseline = dark_baseline['baseline']` is None.

digicampipe/calib/camera/r1.py
dneise: Yes, I think what you are saying makes sense. Just for a performance improvement I wanted to make a minimal change .. since making changes in many places just to improve performance is in my opinion very bad style.

What you propose is to make a change, so that the code is better understandable in the end, and I think that warrants for a change in many places ... 

(To be honest I do not even know in how many places we would have to change something.)

digicampipe/calib/camera/r1.py
dneise: Oh .. right.

dneise:
regardin the performance improvement for tailcut clean, here is the ctapipe PR for your information.
https://github.com/cta-observatory/ctapipe/pull/665

digicampipe/calib/camera/r1.py
dneise: So ... is it okay when we do this change in another PR?

Issue #145, opened 2018-02-28
duplicate?: gain_drop()


dneise:
Similar to #144 I've run into another duplicate, I think:

Here is the 1st:

https://github.com/cta-sst-1m/digicampipe/blob/a7d1f07ab606bf6cce45dc437da39eb851e7549b/digicampipe/instrument/camera.py#L100-L102

Here is the 2nd:

https://github.com/cta-sst-1m/digicampipe/blob/a7d1f07ab606bf6cce45dc437da39eb851e7549b/digicampipe/utils/calib.py#L32-L38

----

These functions are so short, that I can immediately see they do exactly the same, so there is not much harm they can do ... but still this is a duplication, and we should remove one of them.

calispac:
Thank you for pointing that out. The first implementation (in `instruments.camera`) was an attempt to create a sort of `Processor` for the calibration. Where you would init it before looping over the pipeline and it would contain all calibration steps. 
However I think now that this approach is deprecated. Let's keep it simple with functions. 
So I think we can remove what is duplicated in `digicampipe.instrument.camera`. 
I will write a PR for that.

Issue #144, opened 2018-02-28
duplicate: extract_charge()


dneise:
There are two functions with the same name in different modules, which have the same parameters and seem to do the same, or nearly the same.

https://github.com/cta-sst-1m/digicampipe/blob/a7d1f07ab606bf6cce45dc437da39eb851e7549b/digicampipe/utils/utils.py#L27-L48

and 

https://github.com/cta-sst-1m/digicampipe/blob/a7d1f07ab606bf6cce45dc437da39eb851e7549b/digicampipe/instrument/camera.py#L132-L154

I think we should check if they are identical, and if so remove one of them.
If they are not identical but different flavors, I think they should live in the same module but have different names, that underline their differences.

calispac:
See answer in #145

PR #150, opened 2018-03-06
Rm duplicates calib

calispac:
This is a fix for #144 and #145. Here I removed the duplicates functions while checking if they were used.
It turn out nothing was used in the code. 

I am working on the in-lab calibration pipeline and in a future PR I will change the r0-r1 data format. Where it will contain more calibration steps (e.g. the steps in `utils.utils.fake_timing_hist()`)


dneise:
Awesome .. deleting old code is the best that can happen!

PR #151, opened 2018-03-06
New calibration container

calispac:
This is work in progress... In this PR I intend to create special container for the 'in-lab' reconstruction.
Maybe at some point it can be merged with the "data" containers.

For simplification I avoided:
1. data levels (everything is calib)
2. telescope_id (calibration is anyway stopped at telescope level so no need for arrays)

I also took advantage of this PR to: 
* rewrite the reader for digicamtoy (tests still needed).
* add some ressource files for digicamtoy
* add a pixel pulse shape template (pixel 0 only)
* a small example on the Single Photoelectron Spectrum reconstruction (work in progress)

Comments and review are very welcome :smiley:


digicampipe/tests/test_digicamtoy_event_stream.py
```diff
@@ -0,0 +1,59 @@
...
+
+def test_event_source():
+
+    for _ in digicamtoy_event_source(example_file_paths[0]):
+        pass
+
+
+def test_event_stream():
+
+    for _ in event_stream(example_file_paths, mc=True):
```
dneise: the `mc` option was dropped from event_stream. event stream checks the filename to give you the right event_source.

dneise: This is used to guess the event_source

https://github.com/cta-sst-1m/digicampipe/blob/7db37445b6fe32efdc5e5448f8c0cd877b363ee0/digicampipe/io/event_stream.py#L37-L43

digicampipe/utils/utils.py
```diff
@@ -1,9 +1,28 @@
 import numpy as np
 import scipy.ndimage as ndimage
-
+from scipy.interpolate import interp1d
 # Define the integration function
 
 
+def filter_template(filename_pulse_shape, dt):
```
dneise: I do not understand this function name without reading the implementation:
 * is this going to filter some template, as in "I am filtering the template"
 * or is this a kind of getter, which gives me a filter-template, i.e. a template for a filter

digicampipe/utils/utils.py
```diff
@@ -1,9 +1,28 @@
...
+    temp_times = np.arange(time_steps[0], time_steps[-1],
+                           (time_steps[1] - time_steps[0])/1000)
+
+    temp_amplitudes = pulse_template(temp_times)
+    amplitudes = amplitudes / np.max(temp_amplitudes)
+    pulse_template = interp1d(time_steps, amplitudes, kind='cubic',
+                                   bounds_error=False,
+                                   fill_value=0., assume_sorted=True)
+
+    t = np.arange(0, 92*4, 4)
```
dneise: `dt` seems to be unused and these magic numbers should maybe be removed at some place ... 

I like the use of `scipy.interpolate.interp1d` much better than `from scipy.interpolate import splrep, splev` ... *thumbs up*

spe.py
```diff
@@ -0,0 +1,318 @@
...
+
+Options:
+  -h --help     Show this screen.
+'''
+
+from docopt import docopt
+from digicampipe.io.event_stream import event_stream
+import numpy as np
+import matplotlib.pyplot as plt
+from histogram.histogram import Histogram1D
```
dneise: I think this "histogram.histogram" does not exist.

spe.py
```diff
@@ -0,0 +1,318 @@
...
+
+            n_pixels = event.n_pixels
+            adc_histo = Histogram1D(data_shape=(n_pixels, ), bin_edges=np.arange(0, 4095, 1), axis_name='[LSB]')
+
+        adc_histo.fill(event.adc_samples)
+
+    mode = adc_histo.mode()
+    mean = adc_histo.mean()
+    std = adc_histo.std()
+    max = adc_histo.max
```
dneise: I personally would return the `adc_histo` and let the user of this function decide which field of the histogram she might find most interesting.

spe.py
```diff
@@ -0,0 +1,318 @@
...
+
+    return mean, std, mode, max
+
+
+def subtract_baseline(events, value):
+
+    for count, event in enumerate(events):
+
+        event.adc_samples = event.adc_samples - value[:, np.newaxis]
```
dneise: I'd always write this:
```
event.adc_samples = event.adc_samples - value[:, np.newaxis]
```
as:
```
event.adc_samples -= value[:, np.newaxis]
```

Because in the former variant, the reader has to pay close attention if the variable names on both sides of the `=`-sign are the same or not.

spe.py
```diff
@@ -0,0 +1,318 @@
...
+    mean = adc_histo.mean()
+    std = adc_histo.std()
+    max = adc_histo.max
+
+    return mean, std, mode, max
+
+
+def subtract_baseline(events, value):
+
+    for count, event in enumerate(events):
```
dneise: count is unused

digicampipe/utils/utils.py
calispac: This might not end up in PR. I was using different pulse finding algorithm. One that sticked out was : 
find_peak_cwt from scipy (continous wavelet transform.) I was just trying out if I can make a wavelet from the pulse_shape (don't know yet if this is the right approach...). Will work on this soon

digicampipe/tests/test_digicamtoy_event_stream.py
calispac: thanks

spe.py
calispac: correct: at the moment I have it here : https://github.com/calispac/histogram

just a class to create histograms on an "event" basis. I tried to keep it general (i.e. not only for digicampipe) and fast : 

have a look at the example : 
https://github.com/calispac/histogram/blob/master/example.py

For the moment I will add this as a required external packages.

spe.py
calispac: removed

spe.py
dneise: I wrote something similar recently .. maybe we can combine or so.

calispac: yes that would be great!

digicampipe/utils/utils.py
calispac: fixed dt is passed as argument (as well as n_samples i.e. 92)
FYI we usually use 92 instead of 50 samples for in-lab calibration data

spe.py
calispac: ok I tried to do this at some point. This is what I got:

```bash
    event.adc_samples -= baseline[..., np.newaxis]
TypeError: Cannot cast ufunc subtract output from dtype('float32') to dtype('uint16') with casting rule 'same_kind'
```

so in the end I need to convert `event.adc_samples` to float32 to make it work.

dneise: Ah so this means this: 
```
event.adc_samples = event.adc_samples - value[:, np.newaxis]
```
does not only subtract a values from `event.adc_samples` but it also **secretly** changed the dtype of adc_samples. That is quite tricky for a reader to guess at this point.

calispac:
I have no idea why the tests fails (seems to be that the file does not exists ...)

```bash
=================================== FAILURES ===================================
_________________________ test_digicamtoy_event_source _________________________
    def test_digicamtoy_event_source():
    
>       for _ in digicamtoy_event_source(example_file_paths[0]):
digicampipe/tests/test_digicamtoy_event_stream.py:28: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../../miniconda/envs/digicampipe/lib/python3.5/site-packages/digicampipe/io/hdf5.py:43: in digicamtoy_event_source
    hdf5 = h5py.File(url, 'r')
../../../miniconda/envs/digicampipe/lib/python3.5/site-packages/h5py/_hl/files.py:269: in __init__
    fid = make_fid(name, mode, userblock_size, fapl, swmr=swmr)
../../../miniconda/envs/digicampipe/lib/python3.5/site-packages/h5py/_hl/files.py:99: in make_fid
    fid = h5f.open(name, flags, fapl=fapl)
h5py/_objects.pyx:54: in h5py._objects.with_phil.wrapper
    ???
h5py/_objects.pyx:55: in h5py._objects.with_phil.wrapper
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
>   ???
E   OSError: Unable to open file (unable to open file: name = '/home/travis/miniconda/envs/digicampipe/lib/python3.5/site-packages/digicampipe/tests/resources/digicamtoy/test_digicamtoy_0.hdf5', errno = 2, error message = 'No such file or directory', flags = 0, o_flags = 0)
h5py/h5f.pyx:78: OSError
```

the tests are working on my side...

calispac:
All other tests are passing. So I will keep going and fix this later

dneise:
You say at the very top, that this is a special Container for the in-lab reconstruction.

I have absolutely no idea what in-lab reconstruction means, so I find it hard to review. I could only make technical comments maybe... like spe.py seems to need a bit of pep8-ing... but this also a bot can tell you.

Sorry.

dneise:
... I think I start getting a feeling for this PR... maybe no need for you to explain it....

calispac:
I will try to clean it up at the end and add some explanation, I hope it will get clearer. Right now I am experimenting stuff and I see that it is in conflict with data saving. 
> I have absolutely no idea what in-lab reconstruction means

In lab-reconstruction would be camera characterization (in that case SPE analysis). I need to create additional containers because I want the intermediate steps to be saved.

calispac:
I also need to use as much as possible the existing containers because now we have 3 main Containers:
1. Real data
2. MC 
3. Calibration

I think it makes stuff to much confusing

dneise:
Hey @calispac you said, you wanted to split this PR into several parts.... and well ... for me it is easier to review this PR if it was smaller parts. So after todays tele-conference I think I understood more of this PR and feel confident, that I could split it into small parts....

So I'm going to split this PR into little pieces which are easy and quickly to merge... especially those which only add new functionality and do not modify something existing..

dneise:
"how_to_play_with_slow_data.ipynb" was modified in this PR, but not in a meaningful way .. I think ... just some meta data of the notebook was modified ... so I wonder if we should revert that modification, since it is not important.

dneise:
I somehow created a conflict with a test resource ... I do not know how I created that conflict...

calispac:
>  so I wonder if we should revert that modification, since it is not important.

I did not have the intention to change it. So I might have opened it and it changed some meta data. So I think we can revert it

dneise:
Hm ... I must have made a mistake yesterday night ... and since I did not wait for your review ... I merged it into the master. I am sorry for that.

So .. this is the PR where I made a mistake: 
https://github.com/cta-sst-1m/digicampipe/pull/162

I added 12 HDF5 files ... but in this PR .. there is only 1 HDF5 file ... 

I have no idea where I got these 12 files ...

calispac:
It's fine I had 12 files at first : f49acf402dc0ccc31e71c9fbc53660cadf4c0cb0
then removed them

dneise:
Well I still made a mistake .. you added the 12 files on 12.02.18 and yesterday night I added those 12 files to the master ... I was obviously looking at an outdated version of this branch... damn. this should not happen. 

:-(

Still I've got a question regarding `test_digicamtoy_0.hdf5`. In the master (i.e. on 12.02) this file was ~300kB. 
In this PR it is ~5MB. 

So the 5MB file is correct, I assume?

dneise:
In this PR a package was renamed from `calib` to  `reco` or so .. renaming is usually done to improve readability ... mixing this with adding new feature is making it difficult for me to understand.

So .. maybe I just make a new little PR for this renaming right now. Could you tell me why `reco` is a better name than `calib`?

calispac:
> So .. maybe I just make a new little PR for this renaming right now. Could you tell me why reco is a better name than calib?

The current folder structure for low level processing is `calib.camera.r0` and this refers to the processing of the data from waveform to p.e. (IACT real data processing). I wanted to separate: 
1.  "reconstruction" from waveform to hillas
2.  "calibration" from waveform to some SiPM/camera parameters  

Now I realize how it is in this PR is not optimal. maybe : `processing.calib` and `processing.reco` or something like that ?

dneise:
My personal opinion is this: Neither the term "calibration" nor the term "reconstruction" carry enough information about what is going on. So they are all sub-optimal. 

In our field 10 physicists talking about "calibration" mean 23 different things by it. 
["Strictly speaking, the term calibration means just the act of comparison, and does not include any subsequent adjustment."](https://en.wikipedia.org/wiki/Calibration) 
Same is true for "reconstruction". 
[See how many different meanings "reconstruction" can have](https://en.wikipedia.org/wiki/Reconstruction)

So if you say you'd like to rename this from "calib" to "reco" (by the way .. imo the times of short names are over ... we have tab-completion and do not need abbreviations anymore) then I am as unhappy with the naming as I was before. 

So I'd be happy to rename this in one way or another ... just not in this PR.

calispac:
I agree with you on short names it's just as ctapipe did name it : `calib`

dneise:
So .. how should we proceed? remove the renaming from this PR? Or rename calib -> reco in another little PR?

calispac:
> So .. how should we proceed? remove the renaming from this PR? Or rename calib -> reco in another little PR?

Let's forget about renaming in that PR. sorry I should have opened more PRs

environment.yml
```diff
@@ -21,7 +22,7 @@ dependencies:
 - glib=2.53.6
 - gst-plugins-base=1.12.2
 - gstreamer=1.12.2
-- h5py=2.7.1
+- h5py=2.7.0
```
dneise: why do we need an older version of h5py?

calispac: oups sorry forgot to change this again... I was debugging and tried if my issue comes from here

dneise:
Okay nice .. only 18 modified files ... I guess once we hit 10 files or so .. this PR is small enough we can review it as one piece ... 

Sorry, I assume in the end, it turns out this is all good and could have been merged as one piece ... but I really had a hard time wrapping my head around it.. and I assume these tiny PRs we are doing do not hurt too much.

requirements.txt
```diff
@@ -4,3 +4,6 @@ https://github.com/cta-sst-1m/protozfitsreader/archive/v0.44.1.tar.gz
 -e git+https://github.com/cta-observatory/ctapipe-extra.git#egg=ctapipe-extra
 -e git+https://github.com/cta-sst-1m/CTS.git#egg=CTS
 -e git+https://github.com/dstndstn/astrometry.net.git#egg=astrometry
+https://github.com/calispac/histogram/archive/0.1.1.tar.gz
+-e git+https://github.com/iminuit/probfit.git#egg=probfit
+-e git+https://github.com/uvemas/ViTables.git#egg=vitables
```
dneise: vitables seems not to be used anywhere .. do we need it?

pytest.ini
```diff
@@ -2,5 +2,5 @@
 addopts =
     -vv --durations=5
     -k-deselect
-    --benchmark-skip
+     # --benchmark-skip
```
dneise: what do you think, should we keep the benchmarks by default? they seem to take in the order of 5sec all combined ... a bit annoying but still fine...

digicampipe/utils/utils.py
```diff
@@ -1,9 +1,29 @@
 import numpy as np
 import scipy.ndimage as ndimage
-
+from scipy.interpolate import interp1d
 # Define the integration function
 
 
+def filter_template(filename_pulse_shape, n_samples, dt):
```
dneise: This seems to be unused ...

digicampipe/utils/histogram.py
```diff
@@ -0,0 +1,52 @@
+from digicampipe.io.containers_calib import CalibrationHistogramContainer
+import numpy as np
+from histogram.histogram import Histogram1D
+
+
+def convert_histogram_to_container(histogram):
```
dneise: I think this could also be a method of `CalibrationHistogramContainer` ... named `CalibrationHistogramContainer.from_histogram()`.

I mean this function needs to know exactly what the attributes of `CalibrationHistogramContainer` are, so when the container is modified, this function also needs to be modified. It makes sense to keep stuff which needs to be modified for the same reason close together.

digicampipe/utils/histogram.py
```diff
@@ -0,0 +1,52 @@
...
+    container.max = histogram.max
+    container.min = histogram.min
+    container.mode = histogram.mode()
+    container.std = histogram.std()
+    container.mean = histogram.mean()
+
+    return container
+
+
+def convert_container_to_histogram(container):
```
dneise: I think also this could be a method of the container ... `CalibrationHistogramContainer.to_histogram` or `as_histogram`.

dneise: When it is a method of the cointainer, you do not need the type check for example ..

digicampipe/tests/test_imports.py
```diff
@@ -1,13 +1,17 @@
...
+    from digicampipe.utils import geometry
+    from cts_core.camera import Camera
+    from digicampipe.io.save_hillas import save_hillas_parameters
+    from digicampipe.io.save_hillas import save_hillas_parameters
+    from digicampipe.visualization import EventViewer
+    from digicampipe.utils import utils
+    import protozfitsreader
+    from protozfitsreader import rawzfitsreader
+    from protozfitsreader import ZFile
+    from histogram.histogram import Histogram1D
```
dneise: This `test_imports` is from the very very early days, where I got immediately a lot of Exceptions in my fact, when I only wanted to try any example. And the exceptions occured already at the import lines, that is in the very first lines of the script...

Since there were no unit tests what so ever, I thought ... we at least test if the imports work ... at least this.

however this does not mean that we should maintain this test and make sure new stuff is imported here. What we should do is: have at least one unit test per whatever we import.

So instead of checking if `from histogram.histogram import Histogram1D` really works ... why not write a single little unit test which shows in which context Histogram+D is used ... and so on.

Ultimately we have to get rid of this test here and make sure we have real tests ...

mount_sst-1m_data.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+
+# Information can be found here :
+# https://www.tecmint.com/sshfs-mount-remote-linux-filesystem-directory-using-ssh/
+#
+
+sudo sshfs -o allow_other $1@baobab.unige.ch:/sst1m/ /sst1m/
```
dneise: I think this script should be explained in the README otherwise people might never find it ...

dneise:
I think pep8_speaks only updates when the author of the PRs pushes new stuff ...

dneise:
Maybe another comment ... so I did not review spe.py in detail. Since it is a new thing and does not modify anything existing. 

However, I have one general comment here and a few details, which I will comment directly in another review.

pytest.ini
calispac: Yes I think it should be kept. I would say it is important to test if some changes did not affect drastically the computation time.

digicampipe/utils/utils.py
calispac: Thanks

dneise:
Okay ... do you think we should merge this now ... and keep working on spe.py (renaming variables .. maybe speeding up things) in other little PRs?

Merging this gives you a chance to work in the lab, with the master branch, which is always good when results need to be shown.

digicampipe/utils/histogram.py
calispac: Done

requirements.txt
calispac: yes ! do you want to remove it from requirements ?

digicampipe/utils/histogram.py
calispac: Done

digicampipe/tests/test_imports.py
calispac: I just made a test form the container to histogram func. so I can remove it. I will remove the imports that don't need to be testes e.g. protozfitsreader (some tests already perfomed in protozfitsreader package)

requirements.txt
dneise: It is a bit crazy to put recommendations into requirements....

calispac:
I think I added some stuff by mistake in that commit https://github.com/cta-sst-1m/digicampipe/pull/151/commits/958bd9058901b2c87481d3730f4ca63971a02cdc

dneise:
There is a problem with the position of spe.py I think... 
https://travis-ci.org/cta-sst-1m/digicampipe/builds/362135852#L2095

I though I had moved spe.py into `digigcampipe/scripts` to fix this ... but somehow it is back to the root folder.

calispac:
yes I will move it to scripts at the end. actually if it is in `scripts/` I have to install digicampipe each time I make a change to the project. Do you know why ?

dneise:
Thumbs up

PR #165, opened 2018-03-27
add new calib containers

dneise:
**Critical Review needed, do not merge yet**

This is a mini PR from #151. It adds new containers and a new event source, which produces these containers.
.


dneise:
Okay .. so ... I have opened this PR myself ... but now after looking at it, I have a problem with it.

dneise:
So the new `calibration_event_stream` what is it doing? It uses the `event_stream()` to create `DataContainers` which are called `event` inside the function. 

A new `CalibrationContainer` is made ... some part from the new DataContainer is copied into the CalibrationContainer the rest is thrown away. 

Is this really what we want?

dneise:
Or let me phrase the question like this. 

At the moment, the new `calibration_event_stream` is able to real all 3 file-types which are also supported by the underlying `event_stream`: `fits.fz`, `hdf5` and  this 3rd option I always forget about...

Do we want this? Or do we in reality only want to allow hdf5 files for the `calibration_event_stream` ??

calispac:
I mainly created this to throw away the unnecessary `tel` stuff  as well as data levels naming. In any case I think one should revise the low level containers and merge them (simtel, calib, and the `R0CameraContainer`)

calispac:
> Do we want this? Or do we in reality only want to allow hdf5 files for the calibration_event_stream ??

no hdf5 should also be allowed to return a `DataContainer`

dneise:
I think I did not clearly phrase my question:

Should this work? 

    for event in calibration_event_stream('some_fits_file.fits.fz'):
        pass

calispac:
> I think I did not clearly phrase my question:
 Should this work?

yes! I intended to have this `CalibrationContainer` as a simplified container for calibration purpose only (having the minimal structure). Then we can work on making it a unique container for all purposes. Actually I think we could just use the `R0CameraContainer` as a `CalibrationContainer` in the future

dneise:
Okay thanks.

dneise:
Erm .. this test is not passing .. and since an hour I am trying to understand what I messed up.
I mean, this is only a small sub-set of the big PR #151, and I took only a small part from #151 and put it here.
So I wondered how I could have messed up this test in the process...

Only now I looked back to the original #151 and see that the test is also not working there. 
Did it ever pass?

calispac:
Just to summarize it seems that doing this : 
```python

for event_a, event_b in zip(stream_a, stream_b):
    print(event_a, event_b)
``` 

```python
for event_a in stream_a:
    print(event_a)


for event_b in stream_b:
    print(event_b)
``` 

is not printing the same values. In the first example it seems that the two iterators are somehow linked together: if the first increments the second does as well.

PR #168, opened 2018-03-27
Work on digicamtoy event source

dneise:
This is also from  #151.

* Cyril tidied up the `digicamtoy_event_source` and I also tidied it up a bit more.
* Cyril also added some unit tests for the source including benchmarks. (I tidied up the file a little bit)


dneise:
Okay .. tests are passing on my machine .. very nice benchmarks and I like that the digicamtoy source is much shorter now. 

If you are also happy with it, we could merge as soon as travis like it as well.

PR #170, opened 2018-03-28
Spe py in scripts to make it testable

dneise:
This should be merged only after #169 is merged.


calispac:
I cannot merge this to the #169 because tests are failing (since they are missing) I will merge it on my machine and complete the tests

dneise:
Oh .. right ... sorry.

PR #175, opened 2018-04-07
Speed up spe

### LINKED ISSUE #174: digicam-spe does not work: too slow fit_template() and find_pulse_2()
after #172 was fixed I further tried to execute `digicampipe-spe --compute`. I let it run for a while but the progress bar did not move at all.

I put a few `print()` here and there and found that `fit_template()` really took forever. 
So I commented out `fit_template()` and commented in `compute_amplitude()`.
Now it worked .. but it was still very slow.

It measured more of the run times and found that `find_pulse_2()` took 2.2sec per event .. so at 10k events this takes ~22k seconds which is about 6hours ... pretty much exactly what @calispac reported. 

So I played a bit around and implemented a quicker pulse finder, aptly named `find_pulse_3()` .. 
it takes ~1ms per event.

dneise:
This tries to fix #174.

It builds on top of the other PR #173. So whoever reviews this one here, should review maybe #173 first.
Both are pretty short.

### DIFF
--- a/digicampipe/scripts/spe.py
+++ b/digicampipe/scripts/spe.py
@@ -23,9 +23,9 @@ from tqdm import tqdm
 import numpy as np
 import matplotlib.pyplot as plt
 from scipy.signal import find_peaks_cwt
-from scipy import ndimage
 import scipy
 import pandas as pd
+from scipy.ndimage.filters import convolve1d
 
 import peakutils
 from iminuit import Minuit, describe
@@ -206,6 +206,32 @@ def find_pulse_2(events, threshold_sigma, widths, **kwargs):
         yield event
 
 
+def find_pulse_3(events):
+    w = np.array([1, 2, 3, 4, 5, 4, 3, 2, 1], dtype=np.float32)
+    w /= w.sum()
+    SOME_THRESHOLD = 2
+
+    for count, event in enumerate(events):
+        adc_samples = event.data.adc_samples
+        pulse_mask = np.zeros(adc_samples.shape, dtype=np.bool)
+
+        c = convolve1d(
+            input=adc_samples,
+            weights=w,
+            axis=1,
+            mode='constant',
+        )
+        pulse_mask[:, 4:-4] = (
+            (c[:, :-2] <= c[:, 1:-1]) &
+            (c[:, 1:-1] >= c[:, 2:]) &
+            (c[:, 1:-1] > SOME_THRESHOLD)
+        )[:, 3:-3]
+
+        event.data.pulse_mask = pulse_mask
+
+        yield event
+
+
 def compute_charge(events, integral_width, maximum_width=2):
     """
 
@@ -223,7 +249,7 @@ def compute_charge(events, integral_width, maximum_width=2):
         adc_samples = event.data.adc_samples
         pulse_mask = event.data.pulse_mask
 
-        convolved_signal = ndimage.convolve1d(
+        convolved_signal = convolve1d(
             adc_samples,
             np.ones(integral_width),
             axis=-1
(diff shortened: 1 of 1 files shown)


dneise:
Okay I played around with this version myself a bit. I wanted to know if the pulse finder is able to find enough pulses to create a useful charge spectrum.

I played a bit around with `integration_width` and a `shift` (which was a new parameter in charge integration, to try out what happens if we integrate with an offset to where the pulse finder finds the pulse...)

The best combination I found was integration width=5 and shift=0, but also integration_width=4 or 6 or 7 were quite good. I have no idea of the uncertainty of this study. And since somebody already found out that 7 is optimal, I trust them now.

Here is an example spectrum.
spe_spectrum_example

This is a combined spectrum of 10k events and all 1296 pixel. Since in this toy simulation all pixel behave identical, it is fine to add the spectra all together.

dneise:
Ah .. yes .. so the call `digicam-spe --compute` in this call takes ~2min for a file with 10k events and 1296 pixel. About Half of the time goes into the reconstruction of the baseline. But I assume in real-life we would be able to use `digicam_baseline`. I tested this and got ~50sec to `--compute` a 10k events file (single core).

About the pulse finding efficiency one could make this estimation:

 * T = time_per_event x n_events x n_pixel = 4.77 sec
 * NSB rate = 3MHz (dark.yml)
 * --> N_avalanches = 14.3 million

The histogram above contains ~12.5 million entires. So we reach ~87% single pulse finding efficiency with this pulse finder on this toy monte carlo simulation.

**But** .. we might have some double counts. The `pulse_mask` allows for two `True` values in neighboring bins. But the width of a pulse is easily 2 or 3 bins, so two True values next to each other mean: the same pulse was found twice. So in this sense these 87% are only an upper limit. I do not know how many double counts we have.

digicampipe/scripts/spe.py
```diff
@@ -206,6 +206,32 @@ def find_pulse_2(events, threshold_sigma, widths, **kwargs):
         yield event
 
 
+def find_pulse_3(events):
+    w = np.array([1, 2, 3, 4, 5, 4, 3, 2, 1], dtype=np.float32)
```
calispac: Thanks ! this looks pretty good to speed up the pulse finding. if you put `dtype=np.int16` (or the same type as adc_samples) does this speed up even more ?

calispac:
Thanks for the study on peak separation ! Concerning the fit function I think it is great we have it somewhere inside the software. Then we could compare different functions in the future

calispac:
By the way `shift` was introduced to in case the pulse finding algorithm would tend to detect the peak a bit before or after maximum. According to what you have showed it seems that it is not the case but who knows maybe a future peak finding algorithm will have a systematic shift.

calispac:
Maybe to have better dark count rate measurement one could use Poisson count method (i.e. number of zero events). 
In this algorithm the zero are not counted since we explicitly ask for finding pulses. What was told to me by @AndriiNagai is that we could just make a histogram of (`np.max(adc_count) - np.min(adc_count)`).

This will not be so good for gain measurement since the 0. p.e peak is shifted to higher ADC count but would allow us to retrieve dark count rate. Also I doubt that the fit will work with the 0 p.e. peak. 

To create such a histogram as the code is, we need to implement a `find_pulse_x()` that fills `pulse_mask` with `np.arange(0, adc_samples.shape[-1]) == np.argmax(adc_samples, axis=-1)`
and have a function that fills the `baseline` Field with `np.min(adc_samples, axis=-1)`

digicampipe/scripts/spe.py
dneise: I have to try

PR #177, opened 2018-04-08
add shift_parameter

dneise:
I used this parameter in a study I've shown in #175 already. 
So ... shame on me .. I showed results, which nobody can reproduce ... based on code I did not check in.

So here now I propose this version with `shift` in spe.py used by compute_charge

### DIFF
--- a/digicampipe/scripts/spe.py
+++ b/digicampipe/scripts/spe.py
@@ -15,6 +15,10 @@ Options:
   -d --display            Display.
   -v --debug              Enter the debug mode.
   -p --pixel=<PIXEL>      Give a list of pixel IDs.
+  --shift=N               number of bins to shift before integrating
+                          [default: 0].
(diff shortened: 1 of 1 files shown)


digicampipe/scripts/spe.py
```diff
@@ -221,18 +225,18 @@ def find_pulse_3(events):
             axis=1,
             mode='constant',
         )
-        pulse_mask[:, 4:-4] = (
+        pulse_mask[:, 6:-6] = (
```
calispac: why is this smaller now ?

What can you get rid of `[:, 5:-5]` ? or is this method causing issues at boundary ?

dneise: Well ... when integrating with an integration window of size e.g. 7 .. then half of the window is outside the sampling area. So one usually does not even want to consider pulses too close to the border.

PR #179, opened 2018-04-09
use hot-fixed version to read files with missing trigger entries.

dneise:
There is a hotfix now to read zfits files with missing trigger keys.

This is the modification:
https://github.com/cta-sst-1m/protozfitsreader/compare/work_on_v0.44.1_a


calispac:
See my comment to the modified file

dneise:
Thanks for the quick review.

The diff between the original v0.44.1 and the proposed hotfix v0.44.1b0 can be seen here:
https://github.com/cta-sst-1m/protozfitsreader/compare/work_on_v0.44.1_a

PR #181, opened 2018-04-10
remove telescope_id from calibration_event_stream

dneise:
When trying to do `digicam-spe` on observed data I had a problem with the `telescope_id`. Apparently the telescope id in toy-mc files is different from the telescope_id in observations. 

AFAIK we are not interested in setting this number at all, since we **know** in both observations and toy-mc, we only have a single telescope...

So in this PR I remove the explicit value of `telescope_id` and simply use the 1st entry of `event.r0.tel`. 
`tel` is a map (think `dict`) ... but still one can ask a `dict` for a list of all its values and simply take the 1st value. As long as this dict only has a single value anyway, the 1st is always the correct one.

I do not check if there length of the values is exactly one. This means, this will fail as soon as we try this on files with more than one telescope. It might fail silently...meaning that it might simply take one of the many telescopes into account.

I sincerely think this is not a problem inside digicampipe. It would be a problem in ctapipe.

### DIFF
--- a/digicampipe/io/event_stream.py
+++ b/digicampipe/io/event_stream.py
@@ -47,7 +47,7 @@ def event_stream(filelist, source=None, max_events=None, **kwargs):
             yield event
 
 
-def calibration_event_stream(path, telescope_id,
+def calibration_event_stream(path,
                              pixel_id=[...],
                              max_events=None):
     """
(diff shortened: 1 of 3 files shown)


calispac:
Tests are failing because of a typo

Issue #74, opened 2018-01-17
digicam scripts - command line args


dneise:
There are some scripts in `digicampipe/` which cannot be used without modification, as for example they contain fixed paths into somebodies home folder. This should be avoided.

I think it can only be avoided using command line arguments. 
I know about 3 packages which are commonly used for this:
 - [argparse in the stdlib](https://docs.python.org/3/library/argparse.html)
 - [docopt](http://docopt.org/) (also available for C++ as a header-only library, really nice)
 - [click](http://click.pocoo.org/5/)
 
But there is many more:
https://awesome-python.com/#command-line-tools

We should choose one, and use that inside digicampipe.

dneise:
### Input paths

When providing input paths for a program, many ways are possible. In some of the scripts I have seen this option:
```python
    directory = '/home/alispach/data/CRAB_01/'
    filename = directory + 'CRAB_01_0_000.%03d.fits.fz'
    file_list = [filename % number for number in range(3, 23)]
```

One could now think about command line arguments like these:
(using docopt style)
```
Usage:
    scriptname [options]

Options:
-d <dir>, --directory=<dir>  input directory [default:  '/home/alispach/data/CRAB_01/']
--filename=<filename>        template of the filename to be looked for [default:  'CRAB_01_0_000.%03d.fits.fz']
--start=<integer>            first run to be analyzed [default: 3]
--end=<integer>              last run to be analyzed [default: 23]
```

**But** ... this does not follow the linux philosophy. We should not re-implement functionality other people have already implemented for us. In this case it is the bash (or any other shell).

The linux bash (and also the other shells like zsh, tcsh)
is very powerful when it comes to files paths. In fact, this is what it was written for.

So our programs should not reproduce functionality, already implemented very well 
into the shell.

Consider this program call:

    echo foo*.txt

In a folder containing these files:

    foo_1.txt
    foo_2.txt
    foo_3.txt

What does the program `echo` see as its cli arguements? Well, the bash interprets 
the globbing character `*` and invokes `echo` as if the user had typed this:

    echo foo_1.txt foo_2.txt foo_3.txt 

So the user does not have to type in a lot of paths, but the program 
sees a lot of paths. 

Assuming a folder with 100 files with names `foo_000.txt` until `foo_099.txt`
and a user who only wants to "process" 20 files with names from `030` until `049`
might type this:

    echo foo_0[34]?.txt

Which expands to:
    
[...]

PR #183, opened 2018-04-10
run_numbers

jurysek:
Just a small pull request that adding an option for selection of processed data files based on Corsika run number. It is useful when one would like to start with data processing on a subset of competed runs while simulations are still running.


simtel_pipeline.py
```diff
@@ -14,16 +14,24 @@
...
+            yield file
+
+
 if __name__ == '__main__':
 
     parser = OptionParser()
     parser.add_option("-p", "--path", dest="directory",
                       help="directory to data files",
-                      default='../../../sst-1m_simulace/data_test/ryzen_testprod/0.0deg/Data/')
+                      default='/mnt/data1/jurysek/simulace_cta_sst/sim_telarray/cta-prod3-sst-dc-2018/0.0deg/Data/')
```
dneise: how can we avoid putting defaults, that point to some persons private paths. So that our software also works, when a new user wants to try it out?

dneise:
I must say, I do not like this PR. I have written an issue about such user interfaces, here:
https://github.com/cta-sst-1m/digicampipe/issues/74

I think our software should not repeat functionality, which is already contained in normal linux bash.

jurysek:
I agree that while setting a path somewhere to my home directory as a default is very convenient for me, but it is awful for other users :-). You are right that the selection could be done directly from command line with the use of bash tools.

OK then, I will try to rewrite the pipeline with the use of docopt with no default paths.

dneise:
I should maybe soften my statement from before... Also maybe with regard to my new issue #184.

I you personally think you need this feature for perform a certain analysis, then I cannot really say: 
"I do not agree, nobody needs such a feature!"

Just, when we put stuff into the digicampipe, we should somehow have the other people in mind. 

And I can tell you, the first time when I wanted to try a tool written by somebody else, which had a command line interface where I should specify the start-run-number and the stop-run-rumber, this tool exploded right in my face, because I specified start=0, and stop=23, while I only had the runs: 0, 1, 2, and 22, 23 on my disk. 
With this kind of command line interface it was impossible for me to convince the tool to simply process "all the runs I have", simply because there was a gap in run-numbers.

So this command line feature, where the user should specify start and stop run-number, was utterly useless to me as a user. It was possibly useful for the person who implemented it. 

So ... I think the conclusion, to implement minimalistic user interfaces, and then later "glue" the stuff together (using bash scripts or again other python scripts) still holds. It is simply much easier to get a simplistic user interface right.

dneise:
Oh .. you already did the docopt stuff looks good ..!

dneise:
One more comment about the docopt stuff.

I have here an alternative on how to use it ...
https://github.com/cta-sst-1m/digicampipe/compare/hessio_event_source_convenience...explicit_parameters_for_main

In this alternative, `main()` still has a list of explicit parameters ... and down at the bottom these parameters are set from the `args`-dict.

I think the advantage here is this. Assume you want to execute the pipeline, but you want to optimize the cleaning thresholds or so.
In this case you maybe do not want to call it multiple times from the command line, instead you maybe want to write another little script like this:

```python
from simtel_pipeline import main
files = ['some files which are good to study the cleaning']
for picture_threshold in range(10, 20):
    for boundary_threshold in range(5, 10):
        main(
            files,
            outfile_path='.',
            outfile_suffix="pt_{0}_bt_{1}".format(
                picture_threshold,
                boundary_threshold
            ),
            picture_threshold,
            boundary_threshold
        )
```

dneise:
And for even more fun, I have here 
https://github.com/cta-sst-1m/digicampipe/compare/explicit_parameters_for_main...commandr_instead_of_docopt

Compared how the program would look if you'd use `commandr` instead of `docopt`.
`commandr` generates a command line interface from the function signature of the main.
So in my example `main()` looks like this:
```python
def main(
    files,
    outfile_path,
    outfile_suffix,
    picture_threshold=15,
    boundary_threshold=7,
    baseline0=9,
    baseline1=15,
):
```

And from this `commandr` generates this interface ... 
```bash
(digicampipe) dneise@lair:~/sst_new/digicampipe$ ./simtel_pipeline.py 
All options without default values must be specified

Current Options:
 --files=None
 --outfile_path=None
 --outfile_suffix=None
 --picture_threshold=15
 --boundary_threshold=7
 --baseline0=9
 --baseline1=15

No documentation for command 'main'.

Usage: simtel_pipeline.py main [options]
Options without default values MUST be specified

Use: simtel_pipeline.py help [command]
  to see other commands available.

Options:
  -h, --help            
  -f FILES, --files=FILES
  -o OUTFILE_PATH, --outfile-path=OUTFILE_PATH
  -O OUTFILE_SUFFIX, --outfile-suffix=OUTFILE_SUFFIX
  -p PICTURE_THRESHOLD, --picture-threshold=PICTURE_THRESHOLD
                        [default: 15]
  -b BOUNDARY_THRESHOLD, --boundary-threshold=BOUNDARY_THRESHOLD
                        [default: 7]
  -B BASELINE0, --baseline0=BASELINE0
                        [default: 9]
  --baseline1=BASELINE1
                        [default: 15]
```

simtel_pipeline.py
```diff
@@ -195,26 +178,39 @@
...
         n_showers=n_showers,
-        output_filename=directory + hillas_filename)
+        output_filename=args['--outfile_path'] + hillas_filename)
     # save_hillas_parameters_in_text(
     #    data_stream=data_stream,
-    #    output_filename=directory + hillas_filename)
+    #    output_filename=args['outfile_path'] + hillas_filename)
+
+    """
+    # To be added when 'fail_nicely' branch of digicamviewer is in master
```
dneise: Ah .. what is this fail_nicely branch?

jurysek:
Thanks for the review Dominik! I have taken into account the explicit keywords for main(), you are right that is more elegant and useful for testing purposes. Thanks for the tip on commandr, but for now, I will stay with docopt.

jurysek:
Oh.. I am not sure why the check have failed, because the code works fine..

dneise:
> Oh.. I am not sure why the check have failed, because the code works fine..

I restarted the build ... sometimes there is a hickup on travis ...

jurysek:
Great, thanks! So, since it is approved, I am going to merge it into the master.

PR #187, opened 2018-04-13
Remove requirements txt

dneise:
Having an environment.yml and a requirements.txt seems one file too much.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,10 +14,9 @@ install:
 - conda info -a
 - conda env create -f environment.yml --quiet
 - source activate digicampipe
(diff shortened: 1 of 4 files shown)


dneise:
I removed the long running tests again from the travis.yml so we have quicker response times ... do you like that?

dneise:
I think travis sets the environment variable `TRAVIS_EVENT_TYPE` to `cron` when it is a cron job. 
So we could check that one and execute different tests base on this.

https://docs.travis-ci.com/user/cron-jobs/#Detecting-Builds-Triggered-by-Cron

PR #188, opened 2018-04-13
remove file we can import from ctapipe

dneise:
Today @calispac told me that our file `visualization/mpl.py` is in principle copied from ctapipe.

I diffed both files and found no functional difference. So I simply removed our version and imported the version from ctapipe.


dneise:
No no you are totally right. These files were not exactly the same. They were functionally the same. 
in the meantime ctapipe has fixed some pep8 issues in mpl.py which we had not fixed. 

In addition we had an additional parameter named `anti_aliasing` or so ... it was in the docs and in the parameter list ... but it was *not used at all*. 

Anyway ... assume I made a mistake when looking at these files ... and they were really different, I just did not see it. You are right, we do need a test for mpl.py.

A simple first test could be: make a plot .. save it as a file and just check for "no exception" and "file not empty" ... 

Next could be .. store this file in the test resources and make a check that the files are totally identical.

dneise:
I also tested manually that `digicam-view` still works... but I could have forgotten this.

dneise:
Another test would be. Look how we *use* mpl.py .. which functions/classes from mpl.py do we use and how do we use them... then a minimal test is again: make at least sure that we can call these functions or use these classes without throwing an exception.

The next test would include to check that the return values are at least the right type.

Side-effects however .. like making a couple of axis in a figure... are harder to check..

dneise:
Nothing important changed on our side, but the tests fail. I assume something changed in ctapipe. I have no time to investigate that today.

calispac:
If it works then I trust you :D

PR #189, opened 2018-04-17
execute long running tests in daily cron

dneise:
Here I try to execute the long running tests only when `TRAVIS_EVENT_TYPE = cron`, but I am not sure about the conditionals in travis.yml files ... so I will be testing a lot.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,8 @@ install:
 script:
 - pytest digicampipe
 - make -C docs/ html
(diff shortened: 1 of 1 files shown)


dneise:
Seems to work but is ugly ... I used the normal *bash* if and not the *travis* if ... but maybe we can merge this into master and see if it works tomorrow noon for the cron job. What do you think @calispac

dneise:
thanks

PR #192, opened 2018-05-03
Pulse shape.py

dneise:
Proposal for an analysis to find out the average pulse shape generated by the LEDs in the pixels.

by average I mean averaged over many events. Still each pixel get's its own pulse shape.

The analysis takes about 27sec per 10k events.
making the plots takes about 40seconds. 

The result of the "analysis" is simply a histogram with shape: `(*adc_sample.shape, 4096)`, so for example: `(1296, 50, 4096)`. 

dtype is unsigned short, so I just realize, if one wants to histogram many runs with ~10k events per run, one can easily saturate this histogram... I should go for 4byte ints, what do you think?
However, I keep the histogram in main memory. At the moment it is ~500MB. If we go for 4byte ints and maybe for 91 samples or even 300 .. then we need 6GB memory for the histogram alone, maybe this is a problem for some peoples laptops?

Well anyway .. in addition to this histogram, the "profile" is stored as well, so the mean and the std along the last dimension.

Here is an example on how to plot e.g. the profile for pixel 17:
```python
In [1]: import h5py
In [2]: f = h5py.File('SST1M_01_20180502.000.fits.fz.h5')
In [3]: dict(f)
Out[3]: 
{'adc_count_histo': <HDF5 dataset "adc_count_histo": shape (1296, 50, 4096), type "<u2">,
 'adc_count_mean': <HDF5 dataset "adc_count_mean": shape (1296, 50), type "<f4">,
 'adc_count_std': <HDF5 dataset "adc_count_std": shape (1296, 50), type "<f4">}
In [4]: plt.errorbar(
   ...:   x=np.arange(50),
   ...:   y=f['adc_count_mean'][17],
   ...:   yerr=f['adc_count_std'][17])
```
The result could then look like this for example.

figure_1

---- 

in addition one can ask for a lot of plots to be created by the program. It puts them all into one pdf. This plotting takes a while, so when only the histogram (and mean and std) are wanted for further analysis. Then the plotting should be omitted.


dneise:
The tests fail. Something seems to have changed in ctapipe regarding plotting. I have no time to investigate that. It runs on my machine with the version of ctapipe I have installed. 

It seems to be really important to completely update the whole environment regularly ... twice a week or so. @calispac I seem to have problems doing that. Do you know how to really make sure everything is updated? Otherwise I simply completely uninstall and reinstall the entire conda-env twice a week, then I am at least sure everything is up-to-date. 

I just wonder how others do this ...

dneise:
This PR tries to partly solve #193

digicampipe/scripts/pulse_shape.py
```diff
@@ -3,63 +3,211 @@
...
+from docopt import docopt
+import matplotlib.pyplot as plt
+from matplotlib.backends.backend_pdf import PdfPages
+
+from ctapipe.visualization.mpl import CameraDisplay
+from digicampipe.utils import DigiCam
+
+
+def fill_histo_from_buffer(histo, buffer):
+    for pid in tqdm(range(histo.shape[0])):
```
calispac: pid --> pixel_id
sid --> sample_id

also probably better to do:

```python
n_pixels = histo.shape[0]
n_samples = histo.shape[1]
```

before the for loop. I think it makes it more readable

digicampipe/scripts/pulse_shape.py
```diff
@@ -3,63 +3,211 @@
...
+from tqdm import tqdm
+from docopt import docopt
+import matplotlib.pyplot as plt
+from matplotlib.backends.backend_pdf import PdfPages
+
+from ctapipe.visualization.mpl import CameraDisplay
+from digicampipe.utils import DigiCam
+
+
+def fill_histo_from_buffer(histo, buffer):
```
calispac: maybe we can replace this by the `histogram` package. I created it to avoid having to do chunks. I personally think chunks add complexity to the code and are more error prone
As you prefer

digicampipe/scripts/pulse_shape.py
```diff
@@ -3,63 +3,211 @@
...
+        name='adc_count_histo',
+        data=histo,
+        compression="gzip",
+        chunks=(4, histo.shape[1], 128)
+    )
+
+    mean = np.zeros(adc.shape, dtype='f4')
+    std = np.zeros(adc.shape, dtype='f4')
+
+    for pid in range(1296):
```
calispac: let's avoid magic numbers here: we could use n_pixels = adc.shape[0] or len(adc). Also same comment as before in principle one can use the `histogram` package

digicampipe/scripts/pulse_shape.py
```diff
@@ -3,63 +3,211 @@
...
+        )
+
+        pdf.savefig()
+        plt.close('all')
+
+        fig, axes = None, None
+
+        plots_per_page = np.prod(pixel_per_page)
+        print('plots_per_page', plots_per_page)
+        for pixel_id in tqdm(range(1296)):
```
calispac: please let's avoid magical numbers.

digicampipe/scripts/pulse_shape.py
calispac: Ah well maybe I am wrong because you have histogram bin counts of shape = (n_pixels, n_samples)

digicampipe/scripts/pulse_shape.py
dneise: done

digicampipe/scripts/pulse_shape.py
dneise: This needs a 2D histogram, the histogram package at the moment seems to support only 1D histograms

digicampipe/scripts/pulse_shape.py
dneise: thx. done

digicampipe/scripts/pulse_shape.py
dneise: I think a README would help me understanding how to use hat package. 

I tried to understand it from the code, but as it is using C/C++ stuff I was mainly lost after 10 lines...

dneise:
I've reached a new status, one could call ... running.

It is again chunked, which we do not like, but ... well .. it is faster, so I can actually play around...

It can read and histogram the ~160k events of the template scan (see mattermost) in less than 20 minutes on a single core using less than 4GB mem. (NB: These are 32 files with 5k events each, so a single file can be histogrammed in less than a minute. When using a chunk_size of 1000 instead of 3000 events, this needs only ~1.5GB memory, so on my machine for example with 8 cores and 16GB mem in total, it should be possible to run this in less than 4min)

But this is not yet parallelizable easily, since the underlying protozfits reader can only have one file open at a time. The easy way to parallelize is simply to call this script from the shell for every fits file and then combine (simply sum) the resulting h5 files.

dneise:
call it like this:

    digicam-template --output=template_scan_20180515.h5 SST1M_01_20180515_*.fits.fz 

takes ~21min single core (@1.8GB mem now)

digicampipe/scripts/pulse_shape.py
```diff
@@ -3,63 +3,204 @@
...
-        plt.show()
+from tqdm import tqdm
+from docopt import docopt
 
-    pulse_template /= count
+import numba
+
+
+@numba.jit
+def estimate_arrival_time(adc, thr=0.5):
```
calispac: I think this function could be included to this file : https://github.com/cta-sst-1m/digicampipe/blob/master/digicampipe/calib/camera/time.py

where here we try to collect all the time reconstruction methods. These methods :
1. take event containers (calibration events)
2. compute time
3. fill the dedicated field for time

One could rename it : `compute_time_from_leading_edge()`

dneise: Thanks for the hint. You are right. However, I'd like to keep the function interface like this:
```python
def compute_time_from_leading_edge_using_linear(
    adc, # 2d, shape: (n_pixel, n_samples)
    threshold, # 0d, float
):
    # [...]
    return time # 1d, shape:(n_pixel,)
```

and maybe later provide a function which has this interface:
```python
def compute_time_from_leading_edge_using_linear(
    events
    threshold, # 0d, float
):
    # [...]
    yield event
```

The 1st version is simpler to unit test and at least in my opinion simpler to understand... the second version is simply needed in the container-ecosystem. So clearly both versions are needed. The question is just ... how should we do this? 

 - different function names for the different functions? 
 - or different modules? and the same function name?

Dunno ...

dneise:
For a description how to use this PR, please have a look into the README in this PR. 
Also for more details please see the discussion on mattermost.

dneise:
Should we merge this?

calispac:
Yes whenever you feel ready !

PR #195, opened 2018-05-08
pulse template spline interpolation

ImenAlSamarai:
Code for spline interpolation of pulse shapes made by V. Coco.


dneise:
Thank you ... I fixed a few style issues ... now I am going to test it

dneise:
A few thoughts:

 * This program tries to open a file in the current working directory:
```
def estimated_template(pe, start=0, stop=500, step=0.2):
    pkl_file = open('templates_bspline.p', 'rb')
    dict_template = pickle.load(pkl_file)
```

So when I call it like:  `python templates.py` then it works ... but when I am on my Desktop and do: `python path/to/digicampipe/template.py` it does not work. 

This should be avoided.

dneise:
I believe, we can merge this as soon as we put it somehow into context, maybe adding a few lines to the README to explain how to use this and what to expect from this would be nice. 

A more detailed explanation could be put into the module `__doc__` string. 

Once we wrote this, we might be able to think about some unit tests or so ... but at the moment, I believe only a little bit of docu might be needed.

PR #196, opened 2018-05-08
Alternative environment.yml for MAC OS

ImenAlSamarai:
hey @dneise .... this is the env.yml we tqlke about in geneva... hqave a look


dneise:
Hey Imen ... thanks ... I will have a look at it

dneise:
@calispac @ImenAlSamarai 

Hey. As you can see when clicking on the "Show all checks" button above. We have now tests running for Linux and MAC OS, both for Python 3.5 and Python 3.6.

The `environment.yml` is much shorter than before and it does not specify fixed versions for the libraries anymore. Please have a look if you like this.

In principle I think we should specify **minimal versions** not fixed versions as we did before...

So I would not merge this right now ... but I am already quite happy with it anyway.

calispac:
> So I would not merge this right now ... but I am already quite happy with it anyway.

I am confused should we merge ? In my opinion yes

dneise:
> I am confused should we merge ? In my opinion yes

Okay I think you are right ...

PR #197, opened 2018-05-08
Fix bias curve

calispac:
this is fixing an issue with the bias curve.
created an entry point `digicam-rate-scan` and tested it


calispac:
@ImenAlSamarai this is merged into the master

PR #199, opened 2018-05-09
Draw 1296 array to camera

calispac:
this PR includes a function to plot an array of `shape=(1296,)` as a camera view


calispac:
I think this is ready to merge but Travis network seems to be down at the moment...

PR #200, opened 2018-05-11
using latest protozfitsreader version

calispac:
In this PR we try the latest version of the protozfitsreader.

https://github.com/cta-sst-1m/protozfitsreader/archive/1.0.0.tar.gz


calispac:
ok this PR seems to pass the tests. But I doubt that in the newest version it will work it the files do not contain trigger input/output traces. @dneise I will try to fix that in the protozfitsreader

calispac:
Do not merge yet ! this need cta-sst-1m/protozfitsreader/pull/34 to be merged

dneise:
This seems to work for now ... So even if I rather would like to have it in a different way .. should we merge this today so you can work?

dneise:
protozfits v1.0.1 was just relelased:

https://github.com/cta-sst-1m/protozfitsreader/archive/v1.0.1.tar.gz

It is not mentioned on the README yet ... but that's maybe okay.

calispac:
ok thanks!

PR #202, opened 2018-05-12
WIP: Use protozfits v1.0.0

dneise:
This is an alternative to @calispac s #200 I worked on it yesterday on the train, and being offline forgot to push it. Thus causing duplication of work. I am sorry for that... Anyway the difference is this:

At the moment `protozfits` has a thing called `DigicamFile`, which is special both NectarCam and LST use `protozfits.SimpleFile` now. So Regarding the `protozfits` repository my longterm plan is to give the whole repo to the `cta-observatory` organisation. The only reason not to do it yet, is simply because this means giving ownership away entirely ... not sharing ownership with more people. I'd like to share...

Anyway ... I believe we should get rid of the special case `protozfits.DigicamFile`, so in this PR I move stuff from `protozfits.DigicamFile` back here into digigcampipe. (I think last year I made a big deal about moving stuff from digicampipe into the protozfits repo .. and now I am proposing to move it back .. well I guess this simply shows that I am learning ... or that I simply cannot decide what's better.)

This is still work in progress but I had to push it now. 

@calispac please tell me your opinion.


digicampipe/io/zfits.py
```diff
@@ -35,115 +34,173 @@ def zfits_event_source(
...
+                ]
+
+            _sort_ids = None
+
+            for tel_id in data.r0.tels_with_data:
+
+                pixel_ids = event.hiGain.waveforms.pixelsIndices
+                n_pixels = len(pixel_ids)
+                if _sort_ids is None:
+                    _sort_ids = np.argsort(pixel_ids)
```
calispac: I think such sorting should happen in the writer or even before. Main reason being you only what to do this once. But maybe it is for latter.

digicampipe/io/zfits.py
```diff
@@ -35,115 +34,173 @@ def zfits_event_source(
...
+                samples = event.hiGain.waveforms.samples.reshape(n_pixels, -1)
+
+                try:
+                    unsorted_baseline = event.hiGain.waveforms.baselines
+                except AttributeError:
```
calispac: this could be handled at the protzfits package level. as it is for trigger traces

digicampipe/io/zfits.py
```diff
@@ -35,115 +34,173 @@ def zfits_event_source(
...
+                    event.local_time_sec * 1E9 + event.local_time_nanosec
+                )
+                r0.gps_time = (
+                    event.trig.timeSec * 1E9 + event.trig.timeNanoSec
+                )
+                r0.camera_event_type = event.event_type
+                r0.array_event_type = event.eventType
+                r0.adc_samples = samples[_sort_ids]
+
+                r0.trigger_input_traces = _prepare_trigger_input(
```
calispac: we should we prepare the trigger traces again here ? isn't it done in the protozfits package ?

digicampipe/visualization/mpl.py
```diff
@@ -1,584 +0,0 @@
-# Licensed under a 3-clause BSD style license - see LICENSE.rst
-"""
-Visualization routines using matplotlib
-"""
-import copy
-import logging
-
-import matplotlib
-import numpy as np
-from astropy import units as u
```
calispac: This file should not be deleted in this PR

digicampipe/visualization/__init__.py
```diff
@@ -5,7 +5,7 @@
 from matplotlib.widgets import Button, RadioButtons, CheckButtons
 import matplotlib
 
-from . import mpl as visualization
```
calispac: this is not related to this PR

digicampipe/io/zfits.py
```diff
@@ -35,115 +34,173 @@ def zfits_event_source(
...
-        n_events += zfits.numrows
-
-    return n_events
+    return sum(
+        len(SimpleFile(filename).Events)
+        for filename in file_list
+    )
+
+
+PATCH_ID_INPUT = [
```
calispac: if we go for remapping here rather that in the protozfits package then this should be in `utils.Digicam`

digicampipe/io/zfits.py
dneise: My reasoning is: Nothing digicam-specific should be handles in the protozfits package anymore

digicampipe/io/zfits.py
dneise: again: the point of this PR is to remove digicamspecific stuff from the protozfits package and at some point ... please read the text I wrote above in the introduction to this PR

digicampipe/visualization/mpl.py
dneise: Yes ... I messed up my master somehow ... and when I branched this branch off ... this modification was taken also into this branch. ... I have to fix this.

calispac:
Sorry I missed the point of this PR. Now I get it. Anyway remmaping should happen somewhere and to my opinion not in the analysis software nor in the reader but in the writer or the camera itself. So if for now we don't put it as close as possible to the source of the data then having it here or in the protozfits package is the same.

dneise:
> then having it here or in the protozfits package is the same.

I do not agree. I will as soon as possible open a PR in protozfits, which removes everything digicampipe specific from that repo. So the remapping simply cannot happen there. Since it is specific to digicampipe.

calispac:
> I do not agree. I will as soon as possible open a PR in protozfits, which removes everything digicampipe specific from that repo. So the remapping simply cannot happen there. Since it is specific to digicampipe.

I mean it is the same in the sense of the analysis not in the sens you describe. For every event you read (and will probably read it probably more than once) you need to remap.

dneise:
Oh! yes ... it is better to remap only once. But I do not quite see why we are discussing that here. Especially since neither you nor me have a saying in that as far as I understand.

calispac:
> Oh! yes ... it is better to remap only once. But I do not quite see why we are discussing that here. Especially since neither you nor me have a saying in that as far as I understand

Totally agreed ! Anyway with what we have now we need to remap after reading and for me it is an issue and I would have prefer to have this temporary solution as close as possible to the problem. But regarding what you said about specific handling of the files for specific cameras in protozfits package is valid. So let's handle it in this repo until this issue is further discussed.

dneise:
I have to go now ... Can only further work on this tomorrow.

dneise:
This is now trying to use v1.0.2

dneise:
Okay @calispac This would be my current proposal for digicampipe regarding the most recent protozfits release.

dneise:
Event read speed with this seems to be ~189 events per second.

dneise:
Damn I think travis is not a good device for absolute performance tests ... now the results look more like 300 events per sec.

calispac:
- using the current master
- With 50 samples per waveforms
- no trigger traces 
- on a "regular" laptop
- with data copied my laptop SSD disk

I get 500 events per seconds

I will try it with this branch tomorrow

dneise:
Speed seems to be an issue here, so I tried to speed up the protozfits reader, the idea is to do something in the direction of "lazy evaluation". The protozfits v1.0.2 converted every field in a message, which was of type "AnyArray" into a numpy array before handing the message to the user. But if the user only needed access to one of 10 AnyArrays .. then 9 of 10 were converted without need and that took time.

In v1.0.3 the protozfits messages is wrapped in a thin layer, which converts only those parts which are really needed. This saves time.

v1.0.3 is not yet released, but one can test it by installing this branch here:

    pip install https://github.com/cta-sst-1m/protozfitsreader/archive/dynamic_classes.tar.gz

The PR for this branch is here:
https://github.com/cta-sst-1m/protozfitsreader/pull/44

dneise:
Current performance:
```
(foo) ✔ ~/sst_new/digicampipe [use_protozfits_v1.0.0|…28⚑ 1] 
22:19 $ time python test_speed_eventstream.py 
protozfits.__version__: 1.0.3
digicampipe.__version__: 0.1.5
9998it [00:14, 690.48it/s]
events: 9997 duration 14.5
speed: 690.4 evts/sec
```

compared to master-performance:
```
(foo) ✔ ~/sst_new/digicampipe [master ↓·28|…28⚑ 1] 
22:17 $ time python test_speed_eventstream.py 
protozfits.__version__: 1.0.1a0
digicampipe.__version__: 0.1.4
9998it [00:13, 728.55it/s]
events: 9997 duration 13.7
speed: 728.5 evts/sec
```

So this branch is **still** 6% slower than the master branch.

dneise:
Oh ... I tested now also a file **without trigger information** with this branch here (we know the master is a bit faster)
```
(foo) ✔ ~/sst_new/digicampipe [use_protozfits_v1.0.0|…28⚑ 1] 
07:53 $ ipython test_speed_eventstream.py 
protozfits.__version__: 1.0.3
digicampipe.__version__: 0.1.5
0it [00:00, ?it/s]        /home/dneise/sst_new/digicampipe/digicampipe/io/zfits.py:109: UserWarning: trigger_input_traces does not exist: --> nan
  'trigger_input_traces does not exist: --> nan')
/home/dneise/sst_new/digicampipe/digicampipe/io/zfits.py:118: UserWarning: trigger_output_patch7 does not exist: --> nan
  'trigger_output_patch7 does not exist: --> nan')
/home/dneise/sst_new/digicampipe/digicampipe/io/zfits.py:127: UserWarning: trigger_output_patch19 does not exist: --> nan
  'trigger_output_patch19 does not exist: --> nan')
8606it [00:16, 530.24it/s]
events: 8605 duration 16.2
speed: 530.2 evts/sec
```

But isn't that interesting, when there is less data to read, simply because it is missing, the reader is slower.

yrenier:
The number of samples can be arbitrarily chosen up to ~96 if I remember correctly.

dneise:
Thanks @yrenier.

yrenier:
As I get consistently errors at the end of files, I added a `try`

yrenier:
I would like to update the documentation (about which protozfit version to install) and then merge it to master. @dneise is that fine ?

dneise:
Hey @yrenier (i though you are on holiday)
Me an Etienne are just working on yet another version. so maybe do not put any time into it just today.

dneise:
Okay, so I retract the whole lazy evaluation stuff. It was too complicated to hardly any outcome.

So this branch is now using protozfits v1.0.2 again.

dneise:
The issue yves reported about in 
cta-sst-1m/protozfitsreader#48

Was according to Etienne connected to an issue in the rawzfitsreader.cpp.

It had to do with the way Python under the hood deals with return values and exceptions. Something I honestly have no idea about.

Anyway in the new protozfits version we are working on, this issue is fixed by not writing low-level Python-C++ wrappers anymore, but let them be generated by Cython.
This avoids us having to learn all the nasty stuff about Python internals.

dneise:
@yrenier  @calispac should we merge this now?

PR #209, opened 2018-05-24
DISP method

jurysek:
In this pull request, there is a simple implementation of the DISP method for several different disp = f(width/lenght, ..) dependecies.

I decided not to pull some more complicated codes for analysis, like calculations of parameters correlations, investigating resolution dependency on SIZE into digicampipe master, because I think the digicampipe analysis should stay 'basic'. Therefore, the code in this pull request is just for calculation of lookup tables of parameters for chosen equation based on simulations and then image reconstruction (either simulated data for checking performance of the method or real data to see the reconstructed image).

**digicampipe/utils/disp.py**
Contains functions used in analysis, like disp calculation, filling arrival direction matrix or leakage2 calculation. I think that maybe 'leakage2' parameter should be calculated in standard pipeline together with Hillas parameters, because it would save time of loading of large 'events_image*' files for DISP method.

**disp-generate-lookup.py**
Minimization of disp for input simulated data and filling lookup tables for each zenith, azimuth and offset. The first part of the code is sorting of input data. I know that is better to the user to create own sorting procedure in bash or something rather than messing the code with it, but in this case the sorting is based on directory and file structure created by sim_telarray and simtel_pipeline.py so it can be naturaly used. The problem is that sim_telarray creates own directory for each simulated offset and we keep this structure in our database of processed simulations. So, the bash script for data sorting would be long.. If one needs to use all processed files for lookup table creation. Other possibility to avoid this sorting inside the python script is to manually bring all processed files together into one common directory or something like that.. If You have any better idea, please, tell me.

**disp-reconstruct.py**
Just reconstruction of image based on lookup tables. It is possible to switch on the modification of disp method that improves final resolution.


dneise:
I see you use a variable called `method` or `equation` as a kind of switch to perform different operations.
I think due to the fact, that you make decisions based on this value, in different places, this design can lead to problems later when people want to add or remove options.

At the moment, I believe this design affects:
 * `utils/disp.py`: line 13...40
 * `disp-generate-lookup.py`: 114..120 & 164..175 & 189..223 & 227..276
 * `disp-reconstruct.py`: 62..68 & 103..128

If possible, I believe it would be better to find a design, where adding a new equation or removing and old equation needs a change only in one place not in 7 places in 3 files.
(I mean, it is possible, but I do not know if it is possible given the current time constraints)

disp-reconstruct.py
```diff
@@ -0,0 +1,249 @@
...
+  -a <path>, --images=<path>  File with event images.
+  --modification    Turn on modified DISP method.
+'''
+
+import numpy as np
+from docopt import docopt
+import matplotlib.pyplot as plt
+import matplotlib as mpl
+from matplotlib.patches import Circle
+import events_image
```
dneise: `import events_image` only works, when one is in the same folder as `events_image.py` when executing `disp-reconstruct.py` but you cannot make assumptions about any users current working directory (CWD).

disp-reconstruct.py
```diff
@@ -0,0 +1,249 @@
...
+    print(A)
+    (disp_comp, x_source_comp,
+     y_source_comp, residuals) = disp_eval(A, width, length, cog_x, cog_y,
+                                           x_offset, y_offset, psi, skewness,
+                                           size, leakage2, equation)
+
+    # ARRIVAL DIRECTION DISTRIBUTION
+    bins = 100
+
+    if args['--modification']:
```
dneise: `args` is not defined in this function. This is not a syntax error, because `args` is a global variable defined in the `if name == main - block` .... however using global vars is not very good style .. (I do it myself sometimes ... but I think you can simply use `modification` here, no?

digicampipe/utils/disp.py
```diff
@@ -0,0 +1,290 @@
...
+    # theta_squared_cut.
+
+    x_intersect = []
+    y_intersect = []
+    n_bin_values_all = np.zeros((bins, bins))
+    theta_squared_sum_hist = []
+
+    for i in range(len(disp_comp)):
+
+        print(i)
```
dneise: If you want a progress bar, then maybe you'd like to try:
```
from tqdm import tqdm, trange

# for a progress bar, in a for loop using "range", just replace range with trange
for i in trange(len(disp_comp)):
    # do your thing

# for a progress bar, in another loop, use tqdm() around the iterable, so:
for event in events: 
    pass
# becomes:
for event in tqdm(events):
    pass
```

dneise:
General comment: 
I see a lot of for-loops, sometimes 3 nested loops ... this "looks" slow .. if speed is an issue, please let me know, then I can take a closer look at the code with respect to speed, if speed is not an issue, then .. just ignore this comment.

How long does this usually take?

disp-reconstruct.py
jurysek: Right :), corrected.

jurysek:
I do not think that speed is an issue.. There are 3 nested for loops in the _generate_lookup, but I don't think there is a way how to avoid them, since we have to minimize for all combinations of azimuth, zenith, offset. Creation of a lookup table for equation 5 takes ~3 minutes in my laptop for 12 combinations and ~ 25 000 events for each combination.

Then, there are two nested for cycles in arrival_distribution() and it takes also about 3 minutes for 28 000 events with 100 different triplets for each event. Actually, I plan to check dependency of resolution on number of triplets and if turn out that we definitely need more of them, then some optimization would be useful..

jurysek:
I agree with you Dom, that the code is too much dependent on certain numbering of the equations, certain number of parameters and so on, but I have no idea how to make it better.. At least, I got rid of one of those 7 places in disp_generate_lookup.py (189-199) :-)

dneise:
Okay ... so I am now on holiday until next wednesday. So I cannot further look into this. If you think you cannot further improve this, then I guess it is fine enough. 
One can always further improve it later.

dneise:
I think it would be good to present how to use this in the README or so. 

The examples in the doc string are fine .. but it does not explain the whole story...if you know what I mean.
You could for example simply say what you did with it lately... including which files you used and what result one should expect...  So that others can repeat what you did. 

This is particularly imporant if in the future somebody else wants to further improve this code, as long there are no formal tests for this code (and there are absolutely none) ... it is important that others can at least do this:

 * Ah .. I just execute this ... okay .. takes 5 minutes .. 
 * Ah there is a plot as a result ... oh .. 10% DISP resolution ... this is exaclty what Jakub also got back then in 2018 .. so I guess this code still works
 * .. now I will improve the code, so it is easier to read ... it should still do the same afterwards....
 * ... okay let me repeat the analysis ... aha ... 50% DISP resolution ... I guess I fucked it up .. (git checkout previous version)

But for this people need to know the which input to use .. and which output to expect.

jurysek:
OK, thanks. I can wait for a while if Cyril have any comment and then I will merge it.

I plan to describe what I did, together with setup and all plots at sst-1m wiki page. I understand that it would be nice to have some description together with the code somewhere and I can write it too. But in that case, I am not sure where to put it.

dneise:
Ah the wiki is also a good place ...

digicampipe/utils/disp.py
```diff
@@ -0,0 +1,290 @@
...
+import matplotlib.pyplot as plt
+import matplotlib as mpl
+from tqdm import trange
+
+
+def disp_eval(A, width, length, cog_x, cog_y,
+              x_offset, y_offset, psi, skewness,
+              size, leakage2, method):
+
+    parvals = A.valuesdict()
```
calispac: Could you rename `A` to an actual noun please ? I think it makes it clearer what it is

calispac: I would say the same for `parvals`

digicampipe/utils/disp.py
```diff
@@ -0,0 +1,290 @@
...
+
+    x_source_comp0 = cog_x + disp_comp*np.cos(psi)  # Two possible solutions
+    y_source_comp0 = cog_y + disp_comp*np.sin(psi)  #
+    x_source_comp1 = cog_x - disp_comp*np.cos(psi)  #
+    y_source_comp1 = cog_y - disp_comp*np.sin(psi)  #
+
+    # Selection of one specific solution according to skewness
+    # - head of the shower (close to the source) gives more signal
+    # - if skewness > 0, head is closer to the center of FOV than tail
+    x_source_comp = []
```
calispac: would that do the work ?: 

```python
x_source_comp = np.zeros(disp_comp.shape)
y_source_comp = np.zeros(disp_comp.shape)


skewness_is_positive = skewness > 0

x_source_comp[skewness_is_positive] = x_source_comp1
x_source_comp[~skewness_is_positive] = x_source_comp0
y_source_comp[skewness_is_positive] = y_source_comp1
y_source_comp[~skewness_is_positive] = y_source_comp0

residuals = np.array([x_offset-x_source_comp, y_offset-y_source_comp])
```

digicampipe/utils/disp.py
```diff
@@ -0,0 +1,290 @@
...
+    # a simple 1D vector, thus the .flatten() is applied
+    return disp_comp, x_source_comp, y_source_comp, residuals.flatten()
+
+
+def leak_pixels(pix_x, pix_y, image):
+
+    x_rows = np.unique(pix_x)
+    pix_bound = []
+    for x in x_rows:
+        if x < -412 or x > 398:
```
calispac: what are these limitations for ? if they specific to the camera geometry maybe you can use the module : `utils.Digicam`
Which contains information on pixel positions and neighbors. Although I am not totally sure of what is done here.

digicampipe/utils/disp.py
```diff
@@ -0,0 +1,290 @@
...
+
+    # Mask
+    image_mask = np.zeros(len(pix_x), dtype=bool)
+    for i in range(len(pix_x)):
+        for j in range(len(pix_bound)):
+            if pix_x[i] == pix_bound[j, 0] and pix_y[i] == pix_bound[j, 1]:
+                image_mask[i] = True
+
+    # Signal in outermost pixels
+    signal_border = []
```
calispac: I think you can write here  

```python
signal_border = np.sum(image[:, image_mask], axis=1)
```

digicampipe/utils/disp.py
jurysek: Thanks! It works, I just had to use x_source_comp1[skewness_is_positive], because x_source_comp and x_source_comp1 have the same length.

disp_generate_lookup.py
```diff
@@ -0,0 +1,264 @@
...
+    offsets = [x for x in all_offsets]
+    zeniths = []
+    azimuths = []
+    for filename in hillas_files:
+
+        zenith_file = filename.split('_')[-2][2:]
+        azimuth_file = filename.split('_')[-1].split('.')[0][2:]
+
+        if zenith_file not in zeniths:
+            zeniths.append(zenith_file)
```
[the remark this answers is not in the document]
jurysek: OK, thanks for the tip!

disp_generate_lookup.py
```diff
@@ -0,0 +1,264 @@
...
+                     + zenith in x and 'az'
+                     + azimuth in x and offset
+                     + 'deg' in x][0])
+                pix_x = pixels[0, :]
+                pix_y = pixels[1, :]
+
+                min_size = 200
+
+                mask1 = [x > 0.001 for x in hillas['width']/hillas['length']]
+                mask2 = [x > min_size for x in hillas['size']]
```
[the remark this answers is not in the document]
jurysek: Thank You for the tip and for the enthusiastic review Dom, I can't imagine that I would riddle someone's code on a small mobile display :-)

disp_generate_lookup.py
```diff
@@ -0,0 +1,264 @@
...
+
+                # hillas
+                size = hillas['size'][mask]
+                width = hillas['width'][mask]
+                length = hillas['length'][mask]
+                psi = hillas['psi'][mask]
+                cog_x = hillas['cen_x'][mask]  # in mm
+                cog_y = hillas['cen_y'][mask]  #
+                skewness = hillas['skewness'][mask]
+
```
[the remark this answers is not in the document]
jurysek: OK, I added pandas into current commit, but I am sure that I don't use its full power at all.. Maybe I will improve the code later when I get pandas more into my hand.

digicampipe/utils/disp.py
jurysek: This part of the code select the two image pixel rings. So, first it takes X coordinates of the columns and keep two on the left and two on the right. The numbers are X coordinates in mm beyond which the 2+2 columns lie. For upper and bottom side of the camera plane just min and max Y coordinates can be used due to pixel geometry.

disp_generate_lookup.py
```diff
@@ -0,0 +1,264 @@
...
+        for azimuth in azimuths:
+            for offset in offsets:
+
+                print(zenith, azimuth, offset)
+                hillas = np.load([x for x in hillas_files if 'ze'
+                                  + zenith in x and 'az' + azimuth
+                                  in x and offset + 'deg' in x][0])
+                mc0 = np.loadtxt([x for x in pipedmc_files if 'ze'
+                                  + zenith in x and 'az' + azimuth
+                                  in x and offset + 'deg' in x][0])
```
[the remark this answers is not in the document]
jurysek: I see.. I will take a look at it during the weekend and try to figure something out.

digicampipe/utils/disp.py
calispac: This function first tries to find which pixels are at the border of the camera by computing a mask right ?
then computes the ration between "border" image and "full" image to get the ratio of the two --> leakage ?

If so I think one can include the first part into the `utils.Digicam` class. As I think it can be useful in the future. At init of the object `utils.Digicam` ones should fill an attribute called `camera_borders_mask`

calispac: Also if you access the camera object you should be able to get from its attribute `geometry` the neighbor pixels matrix. i.e. a matrix with zeros and ones telling for each pixel which pixel is a neighboring pixel. From that you can count how much neighboring pixels each pixel has. If this count is bellow 6 then it should be at border.

```python
cam = utils.Digicam()
geom = cam.geometry
neighbor_matrix = geom.neighbors
n_neighbors = np.sum(neighbor_maxtrix, axis=0)
camera_border_mask = n_neighbors < 6
```

jurysek: Yes, it is for finding two rows of outermost pixel rings. Thanks for the tip, with the use of camera geometry, it is way more elegant. So, I have just pushed reimplementation with the use of geometry, but including the code into Digicam class is maybe too much for me, because handling with these complicated classes, distributed somehow among multiple files, is a bit beyond my python abilities so far :-)

dneise:
I looked at this again with a fresh set of eyes and realized a fewmore

dneise:
Sorry misclick

jurysek:
Well, I could, because I want to start with ML related stuffs soon since it is widely used in a lot of CTA applications, but not now because I have to also pass some exams in school :-) and I haven't yet any experiences with ML.. 

Actually, I think that ML approach is more general, because one can not only reconstruct shower arrival direction, but also do gamma/hadron separation and probably also energy reconstruction at once, and this is definitely the way how our reconstruction libraries should finally look like. But what I tried to do with my simple 'DISP minimization' approach was just to learn something I didn't know before (before I started read papers as a newbie in gamma astronomy before a half of year, I hadn't believed that image reconstruction with reasonable resolutionwith the use of only one telescope is even possible), and test what resolution we can reach with the use of this old classic method. Then, the next logical step is to improve and generalize the method with the use of huge power of ML and compare its results with the old approach. But now, we have at least some tool, widely used before the ML boom, that we can use for a simple and relatively fast image reconstruction.

dneise:
Oh, I think there is a misunderstanding. I will try to express myself again. I think we crossed our wires about 'Form' and 'Content'. I was talking about form only, while I think you are refering to content.

I believe one can say, that this PR adds 5 (or more?) multivariate Regressors to digicampipe. These regressors are very specialized, since they are meant to be used for DISP regression only, but they are regressors nevertheless. Now my point was this one.

At the moment, the way these regressors are implemented, makes it hard to realize that they are regressors, it took me e.g. 3 days to realize this. 

So I wanted to propose to implement them in a way, which makes it easier for the reader to realize: ah, here is a special regressor being used to predict DISP. 
A homogeneous interface is in my opinion even more important, when people would like to compare different regressors, or even swap two regressors. 

So, trying to stress this point again: I do not want to imply that you should learn about ML and scikitlearn in one night, or even change any of the functionality you have already implemented. I only want to propose to change the interface of the functionality which already exists to an interface which seems to be more common in our community. This interface I am proposing is the interface of a typical scikit learn regressor.

### Time investment .... now?

I am not sure at the moment how much work changing the interface will be. “We do not have much time“ this is always true for everybody of us. Changing an interface **after** everything is 'already finished', feels surely like a waste of time, and if I would now 'request' this change, then it might feel like arbitrary punishment, since everything 'is already done'. 

So in my opinion, the question if it is worth to invest this time now, can be answered by understanding the intention of this pull request. 

**Intention1: Comparison**

Maybe the intention is to implement some disp regressors and compare them with each other. Then include some (already existing) regressors from scikit learn and compare now even more regressors. Then write a paper or a chapter in your thesis about this comparison. 

The result is: You learned a lot about regressors in general and even more about disp regression, abd you can report about this. The code for this comparison is open source and other researchers can repeat this comparison in 5years from now, and it is easy, since all regressors have a very similar interface.


**Intention2: Getting Disp**

If on the other hand, the intention was to just get 'some way' to predict disp from the shower features, then I assume the currently existing 5 equations will in the near future be reduced to only 1. In this case the interface plays absolutely no role, it can behave however it wants. However, we will have a very hard time to prove to others and to ourselves, that our way of predicting DISP is ok. Since we never made a formal comparison, and even if we do a comparison, others have a hard time to repeat it....

---

This all beimg said, I believe we could merge this PR. It adds a festure... and this feature has room for improvement. The same is true for many other features we already have.

But maybe it is not yet the time, to explain this feature to many users and writing a lot of documentation about it, since if/when the interface changes, also a lot of documentation needs to be changed.

jurysek:
Oh, thanks for the explanation. I didn't know that today is so common to use scikit-learn tools for regression, so I missed your point before. Actually, I see that I have probably been living in a lie for ages because I thought that scipy.curve_fit or lmfit.minimize or whatever similar together with definition of some function is still pretty standard way how to 'fit a curve'.. 

I think that the intention of this pull request is somewhere in between of your two proposed. I would like to, of course, make an easily understandable code that can be used for a formal comparison of several regressors. But the original intention at the beginning was probably the second one - to try some old, classic method of image reconstruction and to check whether we really see the Crab or not.

So, I have read yesterday something about scikit-learn and tried to play a bit with LinearRegression model. I think that all the DISP 'equations' could be linearized, but I don't know how to solve the degeneracy that the source can lie on both sides of Hillas ellipse. In addition, it might better to use predicted positions instead of DISP value as 'targets' and use multiple multivariate regression. Or, might be better to create own regressors from scratch, but I haven't found any explanation how to create them that would be clear for me so far.. So, there are many many questions that I have on this scikit approach and I would have to dive deep into all these ML stuff to answer them and to understand how it works and it will take a time and I think I am not able to do it in following days...

dneise:
I am sorry, I am obviously unable to express myself in an understandable way. I never intended you to rewrite your fitting routines using tools from sk-learn. All I wanted to say, is that it might be good to give your **already existing** regressors the same interface as the regressors from sklearn have....

just the interface ....

jurysek:
Sorry, but I am still not sure what exactly you suggest.. You mean that I can create for example a class for each regressor with .fit(), .predict() and so on methods which will contain the minimization procedure from lmfit? If not, could you please give an example?

dneise:
I just realized that the system still was showing, that I request changes... but that was from something older ... 

> You mean that I can create for example a class for each regressor with .fit(), .predict()

Yes exactly.

jurysek:
OK, thanks Dom, I agree with you that scikit-learn like interface seems to be an elegant way how to make the code more clear and prepared for some future extension. But if you don't mind, I would prefer to let it as it is for now and come back to it later.

PR #217, opened 2018-07-09
added local_time to CalibrationContainer

sagreus:
Added local_time to the CalibrationContainer. Please double check for consistency before merge.


calispac:
Thanks!

PR #218, opened 2018-07-16
Fix cron jobs

calispac:
This is an attempt to fix the daily cron jobs:

https://travis-ci.org/cta-sst-1m/digicampipe/jobs/404070843


yrenier:
It seems the astrometry test server is down(I used the test server as it was much faster). I will reconfigure to use the normal one instead.

yrenier:
Passing now on 2/4 machines ! the others failed for time limit reason.
I added the tags to the images uploaded to the astromerty server so we do not need to re-submit them.
As they are already analyzed, it should be faster now.

yrenier:
@calispac passing now :-)

calispac:
Thanks I just changed the `.travis.yml` to do the test for CRON jobs only and added something in the README.md for the MacOS users (not related to this PR but I take advantage of it :D )

calispac:
>>> As they are already analyzed, it should be faster now.
Ah did not see your last comment... 
Should we do these test always now?

yrenier:
The comment was about to go below the 30 min threshold. It still takes >15 min so I would rather  keep it with chron.

PR #222, opened 2018-08-02
fixing slow data

yrenier:
It seems now the slow data fit files are gziped, so few modifications were needed.

@dneise I also changed the set into a list as I had errors there and tried to make the error more explicit.
Let me know if I did something wrong there.

### DIFF
--- a/digicampipe/io/event_stream.py
+++ b/digicampipe/io/event_stream.py
@@ -57,6 +57,7 @@ def event_stream(filelist, source=None, max_events=None, **kwargs):
         except SystemError as e:
             print('WARNING: system error.', e)
 
+
 def calibration_event_stream(path,
                              pixel_id=[...],
                              max_events=None):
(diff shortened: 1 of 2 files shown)


digicampipe/io/event_stream.py
```diff
@@ -42,18 +42,20 @@ def event_stream(filelist, source=None, max_events=None, **kwargs):
...
-            count += 1
-            yield event
+        try:
+            for event in data_stream:
+                if count >= max_events:
+                    raise StopIteration
+                count += 1
+                yield event
+        except EOFError as e:
+            print('WARNING: unexpected end of file.', e)
```
dneise: Instead of print("Warning: ... ") one could use [the warnings package](https://docs.python.org/3.6/library/warnings.html)

yrenier: good point, fixed in next push

PR #225, opened 2018-08-10
fix to plt_rate_vs_threshold

moderski:
Fix to make plt_rate_vs_threshold compatible with rate_scan.py script. Changes: x-axis changed to 'thresholds', plt.yscale fixed, grid added to the plot.


dneise:
We should maybe bump the version a bit before merging

moderski:
I am not quite sure how to bump the version. If you can do it and you think it is necessary - do it. Otherwise, I think it is such a small change that may be merged without changing the version number.

dneise:
you're right: merged

PR #228, opened 2018-08-23
use_protozfits_v.1.4.0

dneise:
In this PR I propose to use the not yet released version of protozfits v.1.4.0.

It is currently under development in a branch named `using_cython_generated_wrapper`, so I just use that branch here.

Since `using_cython_generated_wrapper` does not yet contain the OSX binaries we do expect the OSX tests to fail.

---- 

in the digicampipe code nothing should need to change.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -36,7 +36,7 @@ dependencies:
   - recommonmark==0.4.0
   - secretstorage
   - sphinx-automodapi
-  - https://github.com/cta-sst-1m/protozfitsreader/archive/using_cython_generated_wrapper.tar.gz
+  - https://github.com/cta-sst-1m/protozfitsreader/archive/v1.4.0.tar.gz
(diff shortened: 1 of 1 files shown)


dneise:
Okay just to inform you guys. It is also mentioned in the new README of this branch.

The `Table` class now supports direct random event access. So stuff like this works: `file.Events[1337]` in case Table has enough rows :-D

dneise:
okay protozfits v1.4.0 is out and works.

I assume this can be merged anytime

calispac:
Should we make a new protozfits release to have the OSX binaries? They were updated

dneise:
Ah .. yes .. in fact I will "re-release" v1.4.0 

I must say I was a bit puzzled by the way github handles releases. In fact a `git tag` will trigger a release, even when this tag is not on the master branch.

I assumed a release is only triggered, when the branch is merged into the master.

Issue #230, opened 2018-08-24
Compatibility with ctapipe >= 0.6.0


yrenier:
I got troubles earlier with the pipeline failing because of the "container" keyword to the ctapipe.image.hillas_parameters has apparently been removed in latest release.
Shall we setup a branch to test against new version of ctapipe ?

yrenier:
#231 experience the issue, so no need of a new branch :-)

yrenier:
I ll wait until it is fixed to close the issue

yrenier:
fixed by https://github.com/cta-sst-1m/digicampipe/pull/231/commits/da9ee709b0655f1fb9989165a506c0ae8e20bfde

dneise:
one question: does our pipeline work with ctapipe < 0.6.0 or not? if it does not work with older versions I think there is no need for this if...else ... we just require the version to be > 0.6 and only keep the else part.

We can require versions in the environment.yml or the setup.py or both.,,

yrenier:
I think it works with some versions < 0.6 (for @calispac for example, correct me if I'm wrong).
There must have been some regression which was fixed since, so with the latest version as installed by default by environment.yml is works.
In those conditions I'm not sure what is best

dneise:
okay .. so we keep it like this for the moment ... but in general I think such if version else stuff should be removed asap ... it just makes the code harder to read. 
And honestly, if ctapipe 0.6 is released ... why should we support old stuff and make our live harder?

yrenier:
I agree with you, I will remove it, but I'm not sure how to impose ctapipe version to be >=0.6 in environment.yml as it loading it directly with git.
I also looked in setup.py and it is not mentioned there.

yrenier:
done, cf https://github.com/cta-sst-1m/digicampipe/pull/231/commits/7d7a37c5341329e51e04abf08360c588eccac59f

PR #231, opened 2018-08-28
New pipeline2

dneise:
I think, since this branch is effectively in use now for our online analysis it needs to be reviewed and merged as soon as possible.

So I open this PR.

I do not know if we still need to review the other `new_pipeline` branch, maybe @calispac  or @yrenier can comment on this?


yrenier:
it seems the address of the release of protozfitsreader is wrong: 
`https://github.com/cta-sst-1m/protozfitsreader/archive/using_cython_generated_wrapper.tar.gz` yield error 404.

yrenier:
Also I still have problems with missing columns in the ouptut fits files on baobab, I'm writting test to see if it woks on travis.

yrenier:
about reviewing new_pipeline, as we merged the use_protozfits_v1.0.0 branch into master, there should be no major difference between new_pipeline and new_pipeline2 branch anymore (assuming new_pipeline merged master since)

dneise:
> it seems the address of the release of protozfitsreader is wrong:
https://github.com/cta-sst-1m/protozfitsreader/archive/using_cython_generated_wrapper.tar.gz yield error 404.

The branch has been merged into the master and been deleted. Please use:

    https://github.com/cta-sst-1m/protozfitsreader/archive/v1.4.0.tar.gz

(I'd do it myself, but on mobile I cannot)

dneise:
Cool test!

yrenier:
test with the new adress passes, but I expect it will fail with my new test (I just try on my pc and it failed as well).

yrenier:
We don't get to the part of the test checking the output file, instead the error is related to #230

yrenier:
with latest ctapipe it works (even on baobab) !

dneise:
perfect!

yrenier:
Now I'm fine with merging it, @calispac are you fine with the diff from the new_pipeline branch : https://github.com/cta-sst-1m/digicampipe/compare/new_pipeline...new_pipeline2

calispac:
@yrenier let me have a quick look

dneise:
So the plan is: merge `new_pipeline2` into the master, and simply delete `new_pipeline` and then we have a nice clean state?

digicampipe/tests/test_event_stream.py
```diff
@@ -17,25 +17,3 @@ def test_event_source_new_style():
     for _ in event_stream(example_file_path):
         pass
 
-
```
calispac: These tests should still be performed on `   'example_100_evts.000.fits.fz' ` or can they ?

environment.yml
```diff
@@ -36,10 +37,10 @@ dependencies:
   - recommonmark==0.4.0
   - secretstorage
   - sphinx-automodapi
+  - git+https://github.com/iminuit/probfit.git
```
calispac: This can be removed

yrenier: it is at least used in validate_trigger_digicam.ipynb

yrenier: also in:
digicampipe/calib/camera/charge.py
build/lib/digicampipe/scripts/fmpe.py
build/lib/digicampipe/scripts/spe.py

calispac: ok then let's keep it. In the long term I want to avoid probfit. In my experience it is slowing down computation

digicampipe/tests/test_event_stream.py
yrenier: Tests are done on both.
It is my bad: I copied digicampipe/tests/test_event_stream.py to digicampipe/tests/test_event_stream2.py and modified the original.

yrenier:
deleting branch as https://github.com/cta-sst-1m/monitoring_pipeline was updated to use master

PR #233, opened 2018-09-05
fix precission problem in ns timestamps

yrenier:
if one do:
```
from digicampipe.io.event_stream import event_stream
events = event_stream('/sst1m/raw/2018/09/01/SST1M_01/SST1M_01_20180901_196.fits.fz')
event = next(events)
tel = event.r0.tels_with_data[0]
t = event.r0.tel[tel].local_camera_clock
print(t - (t+1))
```
with the current code, it prints 0 !
That's because local_camera_clock is defined as a float and floats does not seems to have 10^18 precision.
Using integers instead fixes the problem.

### DIFF
--- a/digicampipe/io/containers.py
+++ b/digicampipe/io/containers.py
@@ -22,6 +22,7 @@ import pickle
 from os.path import isfile
 from ctapipe.io.serializer import Serializer
 from os import remove
+import numpy as np
 
 __all__ = ['InstrumentContainer',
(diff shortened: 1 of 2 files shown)


yrenier:
It was the reason Matthieu could not find trigger with less than 256 ns between them.
in the code below `print(t - (t+128))` prints -256 !

calispac:
Ok nice !

PR #234, opened 2018-09-07
Fix pipeline parameters

calispac:
In this PR the charge to amplitude ratio is computed from the pulse template. I also corrected input of gain in computation of nsb.


environment.yml
```diff
@@ -39,8 +39,8 @@ dependencies:
...
   - https://github.com/cta-sst-1m/protozfitsreader/archive/v1.4.0.tar.gz
-  - git+https://github.com/cta-observatory/ctapipe.git#egg=ctapipe
-  - git+https://github.com/cta-sst-1m/CTS.git#egg=CTS
-  - git+https://github.com/dstndstn/astrometry.net.git#egg=astrometry
+  - git+https://github.com/cta-observatory/ctapipe.git
+  - git+https://github.com/cta-sst-1m/CTS.git
+  - git+https://github.com/dstndstn/astrometry.net.git
   - https://github.com/calispac/histogram/archive/0.3.2.tar.gz
-  - git+https://github.com/uvemas/ViTables.git#egg=vitables
+  - git+https://github.com/uvemas/ViTables.git
```
yrenier: What was that `#egg` argument doing ?

run_pipeline.sh
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
 
-digicam-raw --compute --output=./ /sst1m/raw/2018/08/19/SST1M_01/SST1M_01_20180819_{001..003}.fits.fz
-# digicam-pipeline --compute --display --output=. --dark=raw_histo.pk --parameters=digicampipe/tests/resources/calibration_20180814.yml /sst1m/raw/2018/08/19/SST1M_01/SST1M_01_20180819_{004..010}.fits.fz
\ No newline at end of file
+# digicam-raw --compute --output=./ /sst1m/raw/2018/08/19/SST1M_01/SST1M_01_20180819_{001..003}.fits.fz
+digicam-pipeline --compute --display --output=. --dark=raw_histo.pk --parameters=digicampipe/tests/resources/calibration_20180814.yml /sst1m/raw/2018/08/19/SST1M_01/SST1M_01_20180819_{009..010}.fits.fz --template=digicampipe/tests/resources/pulse_SST-1M_pixel_0.dat --debug
```
yrenier: for the run_pipeline.sh script, I would let both digicam-raw and digicam-pipeline as both are needed when working from raw data.

calispac: Ok I change the script. Now I will not update it anymore

calispac:
I am working on a class `PulseTemplate` to handle the calculations that are done at the beginning.
It could in the future also be used to call a template function e.g. (does the interpolation between samples)

```python
template = PulseTemplate.load(pulse_template_filename)
time = np.linspace(0, 200, num=100000)
template(time, amplitude=3, t_0=10)
```

environment.yml
yrenier: fom pip --help:
  --egg                       Install packages as eggs, not 'flat', like pip normally does. This option is not about installing *from* eggs. (WARNING: Because this option overrides pip's normal install logic, requirements files may not
                              behave as expected.)

setup.py
```diff
@@ -25,15 +25,14 @@
...
         'numpy',
         'matplotlib',
         'scipy',
         'astropy',
         'h5py',
         'tqdm',
     ],
-    tests_require=['pytest>=3.0.0'],
+    tests_require=['pytest>=3.0.0', 'ctapipe>=0.6.0'],
     setup_requires=['pytest-runner'],
```
yrenier: from https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-dependencies
it seems `install_requires` is the right place, `tests_require` is for package only needed during tests.

calispac: Ok changing this

environment.yml
calispac: So it is good that we changed this ?

yrenier: I am not sure, my guess is that it should not have much impact.

calispac:
Now I see why the PR is not passing. The example file is not a file taken during observations and the pipeline applies cuts to cleaned spots that  touch the border. Here is an example off event 
image

I will make the pipeline not discard events that touch the camera border but rather flags it

calispac:
@yrenier could you review ?

yrenier:
I will review after the meeting this afternoon.

digicampipe/calib/camera/cleaning.py
```diff
@@ -68,8 +68,13 @@ def compute_boarder_cleaning(events, geom, boundary_threshold):
...
+        event.data.border = on_border
 
-        if not on_border:
+        if not on_border and skip:
+
+            yield event
+
+        else:
 
             yield event
```
yrenier: I guess it should be:
```
        if on_border and skip:
            continue
        else:
            yield event
```

calispac: Thanks

Issue #236, opened 2018-09-08
digicam baseline in digicam-raw script


moderski:
It would be useful to have a digicam calculated baselines also computed by the digicam-raw script (unless it is already there, and in such a case it would be nice to have histograms of baselines calculated as an average of the signal within readout window).

PR #237, opened 2018-09-10
Baseline in digicamraw

calispac:
Working on #236

### DIFF
--- a/digicampipe/scripts/raw.py
+++ b/digicampipe/scripts/raw.py
@@ -25,6 +25,7 @@ from histogram.histogram import Histogram1D
 from digicampipe.io.event_stream import calibration_event_stream
 from digicampipe.utils.docopt import convert_max_events_args,\
     convert_pixel_args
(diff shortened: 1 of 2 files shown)


calispac:
Here are some images of the display. This was not done on dark data
image
image

digicampipe/scripts/raw.py
```diff
@@ -9,12 +9,16 @@
...
   -o FILE --output=FILE.      File where to store the results.
-                              [Default: .]
-  -c --compute                Compute the data.
+                              [Default: ./raw_histo.pk]
+  -c --compute                Compute the raw data histograms.
   -d --display                Display.
   -p --pixel=<PIXEL>          Give a list of pixel IDs.
   --save_figures              Save the plots to the same folder as output file.
-'''
+  --baseline_filename=FILE    Output path for baseline histogram. If None the
```
moderski: It is not clear what baseline histogram it is. I would suggest something like "Output path for DigiCam calculated baseline histogram."

PR #239, opened 2018-09-10
Sphinx doc update

calispac:
Trying to complete the doc and use the autoapi package for shinx.


calispac:
Yes we can merge anytime in principle but. It is not a complete documentation at the moment. I am trying to make the "auto-magic" documentation with sphinx. I.e. one page of documentation per file/module with the doc string of each function.

PR #240, opened 2018-09-11
Event ID

calispac:
Access events by event_ID.


calispac:
@vic0ward if you need to start the event stream by an given event_id

Issue #227, opened 2018-08-20
pyhessio issues on CentOS6


dneise:
@Etienne12345 told me pyhessio does not work on SL6. 

We are unfortunately importing pyhessio by default in [`event_stream.py`](https://github.com/cta-sst-1m/digicampipe/blob/master/digicampipe/io/event_stream.py#L1)

I believe if pyhession is not installed we throw an exception, even if the user will never even try to read a hessio file.

It would be better to import this stuff only when really needed.  ctapipe is adopting the same strategy here: only import if really needed.

Should be easy to fix, just writing it here to not forget about it.

calispac:
That sounds good enough. @jurysek do you have an opinion? since you are using pyhessio

dneise:
What I have in mind should not affect @jurysek in any way. He will not realize that the import is only done, when he actually tries to read the first file instead of doing it on first import.

Also runtime should not be negatively affected.

calispac:
What I realized on "baobab" is that the installation of pyhessio from the anaconda repository did not work. In fact it was better to git clone pyhessio and then python setup.py install. Which is what was corrected in #241 . I think the difference is that C library pyhessio files are compiled with setup.py install which is then compiled with the GCC from the user machine. Did not really check but it worked for me

calispac:
Baobab uses CentOS release 6.10 (Final)

PR #241, opened 2018-09-11
installing `pyhessio` for the source directory. therefore c libraries ar…

calispac:
…e compiled on the user machine


calispac:
#227

PR #242, opened 2018-09-12
Project cleanup

calispac:
As mentioned in the title this is a project cleanup. I replaced files where they should be and removed duplicate code. Do not merge yet. 

As this is quite a large PR I would appreciate if developers could tell me if I broke something or if you still have what you need. 

WIP


calispac:
Once the tests are passing I will do some pep8

yrenier:
Thanks for this, that was needed :)

yrenier:
I'm not a big fan of "personal" folders in projects, @jurysek it would be nice if you can come up with a folder name a bit more descriptive than jakub.

calispac:
I putted the scripts of @jurysek into the jakub/ folder. I think we should put some off the functions into the digicampipe package. I cannot guaranty that I did not destroy it. But anyway I think we should have tested more functions in this project. Therefore we cannot support code that is not tested. 
Could be good if we dedicate an other PR in the future to tests every single function

calispac:
> I'm not a big fan of "personal" folders in projects, @jurysek it would be nice if you can come up with a folder name a bit more descriptive than jakub.

My bad I just renamed it like this. I did not know exactly were to put what

calispac:
@jurysek thanks for the info I will change what you suggested

calispac:
What about mc_shower.py (seems to save a container to a txt file which we can do easily with fits files now)  and energy-generate-lookup.py (I put it in scripts/) ?

In simtel_baseline I see baseline computation methods I moved to baseline.py and adapted it.

dneise:
Great Work!

jurysek:
energy-generate-lookup.py can be also moved to scripts/. mc_shower.py can be deleted together with simtel_pipeline. Could you point me to the functions which I can use for saving MC parameters into fits files? I will have to also rewrite the analysis scripts a bit in order not to load parameters from those txt files any more.

calispac:
Sure ! you can find an example in scrpits/pipeline.py. In few steps:
1. Need to create a container with what you want to save
2. Then create a `ctapipe.io.serializer.Serializer`
3. While looping over the events fill the container 
4. Call my_serializer.add_container(you_container)
5. At the end of the loop close serializer

dneise:
There are many files in scripts/. which are not (yet?) listed in the `console_scripts` in setup.py.

I think it could help users if there was a simple rule of thumb, like:
> a script named e.g.  `scripts/foo.py` can always be called as `digicam-foo` from the command line.

We only need this, I think:
```python

def make_console_scripts():
    import glob
    import os.path

    scripts = []
    for path in glob('digicampipe/scripts/*.py'):
        filename = os.path.basename(path)
        stub = filename.split('.')[0]

        # this way we can "hide" a file in scripts/
        # from becoming a console_script
        # and we naturally ignore __init__.py if it exists.
        if filename.startswith('__'):
            continue

        scripts.append(
            'digicam-{stub}=digicampipe.scripts.{stub}:entry'.format(
                stub=stub
            )
        )
    return scripts


setup(
    name='digicampipe',
    # [...]
    entry_points={
        'console_scripts': make_console_scripts(),
    }
)
```

dneise:
Or shorter if you like :-D
```python
def make_console_scripts(glob_expr='digicampipe/scripts/*.py'):
    import glob
    import os.path

    return [
        'digicam-{stub}=digicampipe.scripts.{stub}:entry'.format(
            stub=os.path.basename(path).split('.')[0]
        )
        for path in glob(glob_expr)
        if not os.path.basename(path).startswith('__')
    ]
```

calispac:
@dneise Nice ! I think the only problem is that not all scripts are tested and working so for the moment can be better to do it manually. But maybe it does not hurt.

dneise:
up to you .. we can also do it later... but I think the earlier the better ... I have the impression, if a script is not tested or maybe not working then having some people stumble over this and ask why something is not working .... can actually help :-D.

Anyway the code above has bugs. I am currently debugging it and when it works I'll put it in a gist and link it here or so.

calispac:
Ok sounds good! let's put this auto console script feature

dneise:
> Actually if the script does not have an entry() function defined then it does not make a command line app.

I do not think so. The command line thing exists, but it simply fails .. just as you said before.

> Therefore we can put "development" scripts in in scripts/ without having it as a cline app.

People who know, that they just put something in `scripts/` which does not work, can "hide" it by renaming it like `scripts/__something.py` and only when they feel ready, they rename it to `scripts/something.py`

dneise:
ed86c63 renamed  `digicampipe.scripts.data_quality.entry` to  `digicampipe.scripts.data_quality.main`.
This made a test fail. I modified the test, to test main() instead of entry(). 

It seems `digicampipe.scripts.data_quality.entry` was not used anywhere else, so this should be fine now.

yrenier:
filter is an existing python function (https://www.w3schools.com/python/ref_func_filter.asp), better to rename that module

Issue #244, opened 2018-09-27
event_type determination should be done bitwise


moderski:
Muon flagging on camera server is going to be implemented. It is assumed that the algorithm will set new bits in camera_event_type field:
```
#define EVTYPE_CSP_MUON         (0x10000)       /* camera server detected muon*/
#define EVTYPE_CSP_HILLAS       (0x20000)       /* camera server computed Hillas parametrs */
```
additionally to already existing set of bits:
```
#define EVTYPE_DIGICAM_ALG0     (0x1)           /* algorithm 0 trigger - PATCH7 */
#define EVTYPE_DIGICAM_ALG1     (0x2)           /* algorithm 1 trigger - PATCH19 */
#define EVTYPE_DIGICAM_ALG2     (0x4)           /* algorithm 2 trigger - MUON */
#define EVTYPE_DIGICAM_INTRNL   (0x8)           /* internal or external  */
#define EVTYPE_DIGICAM_EXTMSTR  (0x10)          /* unused (0) / external (on master only)  */
#define EVTYPE_DIGICAM_BIT5     (0x20)          /* unused (0)  */
#define EVTYPE_DIGICAM_BIT6     (0x40)          /* unused (0)  */
#define EVTYPE_DIGICAM_CONT     (0x80)          /* continuous readout marker  */
```

Since many event_type assessments in digicampipe are done numberwise `if event_type == 1:` they should be changed to bitwise `if event_type & 0x1:`

calispac:
Ok good point ! After some offline discussion we thought we could maybe use the `in` statement in python.

e.g. `if event_type.CSP_MUON in event_type`. To do this one would need to create a class which has a function `__contains__()` defined.

dneise:
I looked further and found a nice solution for this in the Python stdlib module `enum`. The class doing this is called `IntFlag` 

So I propose this class:
```{python}
class CameraEventType(enum.IntFlag):
    '''
    I removed the EVTTYPE, since this is clear from the class name
    I removed DIGICAM since this is clear from the project.
    '''
    ALG0 = 0x1
    ALG1 = 0x2
    ALG2 = 0x4
    INTRNL = 0x8
    EXTMSTR = 0x10
    BIT5 = 0x20
    BIT6 = 0x40
    CONT = 0x80
    CSP_MUON = 0x10000
    CSP_HILLAS = 0x20000
```

It can be used like this: ...

dneise:
[...]

PR #245, opened 2018-09-27
Camera event type contains

dneise:
Proposal regarding issue #244 
@moderski et all. please look how you like this.

Please especially keep an eye on the fact, that assignment: name <-> numerical_value for the bit fields of the enum come out of nothing. That is not ideal. I would like to improve this, but I do not know exactly how.


calispac:
Apparently tests are failing on python 3.5

dneise:
Ah right ... sorry ... I forgot it over the weekend .. thanks for reminding ... gonna look at it now.

dneise:
okay seems to be running now

yrenier:
It's great ! Should we keep the commented parts in case we drop 3.5 compatibility or get rid of them ?

dneise:
doesn't matter .. once we drop py3.5 and support only 3.6 and 3.7  we can get rid of the entire `enum_flags.py` file ...

dneise:
We should never really look into `enum_flags.py` ... just ignore it and assume it works like the enum classes which exist in Py3.6 ... even though we night not get exactly the same performance in py3.5.

dneise:
I can remove the comments. wait

yrenier:
I also found event_type used in: 
- digicampipe/io/__init__.py at line 80 `if self.r0_container.camera_event_type is not None:`
- fill_event_types() in digicampipe/calib/filters.py
- filter_event_types() in digicampipe/calib/camera/filters.py
- fill_baseline_r0() in digicampipe/calib/baseline.py

dneise:
But the tests are passing.

dneise:
`digicampipe/io/__init__.py` is an empty file

yrenier:
sorry it was digicampipe/visualization/__init__.py

dneise:
Okay anyway .. we know there is code in digicampipe, which was written .. then used ... then not used anymore but not deleted. 

So: passing tests might mean, the functions you found are simply dead code, which should not be fixed but deleted.

Or it means these functions are indeed important, then we should write tests for them. But I cannot say if something is important at the moment.

digicampipe/io/containers.py
```diff
@@ -44,6 +47,38 @@
...
+    BIT5       (0x20)    /* unused (0)  */
+    BIT6       (0x40)    /* unused (0)  */
+    CONT       (0x80)    /* continuous readout marker  */
+    CSP_MUON   (0x10000) /* camera server detected muon*/
+    CSP_HILLAS (0x20000) /* camera server computed Hillas parametrs */
+
+    I removed the EVTTYPE, since this is clear from the class name
+    I removed DIGICAM since this is clear from the project.
+    '''
+    UNKNOWN = 0x0
```
calispac: I wonder if we should use the same naming as done in other code.  I would prefer that names are words. e.g. : 

ALG0 --> ALGORITHM0
INTRNL --> INTERNAL
EXTMSTR --> ??? I don't even know what it is
CONT --> ???
CSP_MUON -- ??? is it some king of cameraserverflag?

digicampipe/io/containers.py
```diff
@@ -156,7 +191,16 @@ class R0CameraContainer(Container):
     local_camera_clock = Field(np.int64, "camera timestamp")
     gps_time = Field(np.int64, "gps timestamp")
     white_rabbit_time = Field(float, "precise white rabbit based timestamp")
-    camera_event_type = Field(int, "camera event type")
+    _camera_event_type = Field(CameraEventType, "camera event type")
+
+    @property
```
calispac: I don't understand why we have 2 "camera_event_type"s one is "private" the other is "public".
If only 1 is required I would keep the "public" one

digicampipe/io/containers.py
```diff
@@ -44,6 +47,38 @@
...
+    ALG0 = 0x1
+    ALG1 = 0x2
+    ALG2 = 0x4
+    INTRNL = 0x8
+    EXTMSTR = 0x10
+    BIT5 = 0x20
+    BIT6 = 0x40
+    CONT = 0x80
+    CSP_MUON = 0x10000
+    CSP_HILLAS = 0x20000
```
dneise: do we really like these names? `INTRNL`??? really? I assumed someone would speak up here.

at least I can tell you that I misspelled this variable already like:
 - INTERNL
 - INTRNAL
 - INTERNAL

because it takes actually longer to remember which letters had been removed than typing the correctly spelled english word internal.

calispac: See my comment

digicampipe/io/containers.py
dneise: what do you mean by two? the red line was removed ... the green line now exists... the public was converted into a private and needs to go through a setter and getter now, so the type is correctly converted.

digicampipe/io/containers.py
calispac: ALG0 --> CLUSTER7
ALG1 --> CLUSTER19
ALG2 --> MUON
INTRNL --> INTERNAL
BIT5 --> No clue what it is
BIT6 --> No clue what it is
CONT --> No clue what it is
CSP_MUON --> No exact clue what it is
CSP_HILLAS --> No exact clue what it is

digicampipe/io/containers.py
dneise: I agree, but if *we* as SST1M digicam people now decide we want to rename these. Do we ask the camera server people to also rename? Or do we create another wall of mental mapping?

digicampipe/io/containers.py
dneise: > Sorry to ask again 
did you ask that already before? ... see it the first time... maybe... I missed it before sorry.

Good question... maybe ... In fact, maybe you want to have a look at this from the CamerasToACTL repo:

https://forge.in2p3.fr/projects/ctaactl/repository/entry/CamerasToACTL/trunk/DataModel/L0.proto#L207

The File `L0.proto` contains the protobuf data model, which describes how the for example the `CameraEvent` looks like (line 329 of L0.proto). The link above goes to line 207 of the file, where an enumeration is defined for exactly what we are discussing here.

We observe two things immediately:

 * SST1M does not use this.
 * This enum does not define a bitmap encoding, where and event can be ALG0 **and** CONT at the same time, but here the type is either PHYSICAL **or** PEDESTAL.

If we look deeper into it, we see the enum from line 207 named `EventType` is used for the field named `eventType` defined in line 333. Which field are we using then?? Ah of course .. we are using the field `event_type` which is not an enum, but a simple integer...why? Well that's obvious, because the `EventType` enum was not suitable for our use case...

Anyway somewhere down the line people took decisions.... and now we (at pretty much the end of the line) have to live with it... but asking us to live with it in clearly readable code, which is easy to understand? Well I guess that is not so easy, since the whole problem is not easy to understand in the first place.

yrenier:
Also it seems the enum_flags.py does not work with my python 3.5.4, I get:
```  File "/home/yves/ctasoft/digicampipe/digicampipe/io/enum_flags.py", line 4, in <module>
    class Flag(Enum):
  File "/home/yves/ctasoft/digicampipe/digicampipe/io/enum_flags.py", line 7, in Flag
    def _generate_next_value_(name, start, count, last_values):
  File "/home/yves/anaconda3/envs/digicampipe/lib/python3.5/enum.py", line 61, in __setitem__
    raise ValueError('_names_ are reserved for future Enum use')
ValueError: _names_ are reserved for future Enum use```

**Comment by dneise** [comment]
Okay so then, I guess we should stop the `enum.Flag` experiment here. We can still do it the way Rafal proposed: Let the type stay an integer and do simple bitwise comparisons...

Who wants to implement that?

**Comment by yrenier** [comment]
I don't know, it looked so much better that the integer :-(

**Comment by yrenier** [comment]
i found [aenum](https://pypi.org/project/aenum/) which is compatible back to python 2.7 and has a very similar IntFlag class

**Comment by dneise** [comment]
perfect! let's try that!

**Comment by dneise** [comment]
you or me?

File: digicampipe/io/containers.py [hunk 4]
```diff
@@ -484,6 +528,6 @@ class CalibrationContainer(Container):
     pixel_id = Field(ndarray, 'pixel ids')
     data = CalibrationEventContainer()
     event_id = Field(int, 'event_id')
-    event_type = Field(int, 'Event type')
+    event_type = Field(CameraEventType, 'Event type')
```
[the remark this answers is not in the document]
dneise: Ah! well this is a bug, I think. Well spotted.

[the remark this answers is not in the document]
dneise: > What is the actual difference between the two? 

I remember I asked you this question months ago... I guess we both do not know it, hu?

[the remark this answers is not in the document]
dneise: This is not about public or private .. it is about .. do we need a setter and getter to convert the type from int to something special or not? .. and I believe here we need a setter and getter again...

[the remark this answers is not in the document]
dneise: Ah no ... I am confused .. I did not ask you month ago about these ...

yrenier:
I just sent an e-mail around to rename those flags

yrenier:
@dneise don't give up, you did all the work !

digicampipe/io/containers.py
calispac: if we don't use `event_type` shouldn't we just remove it from the container?

dneise:
@yrenier looks beautiful so far... 
We could merge it, so that Rafals issue is solved for the time beeing and improve the event type names later. 
(It is always a pity, when something that works is not merged, because we are waiting for a details, which can easily be added also in a dedicated PR later)

dneise:
Oh do we need to update the version number ?

yrenier:
v0.2.1 ?

digicampipe/io/containers.py
yrenier: from Jurek:
CSP_HILLAS flag is set whenever:
* NO CSP_MUON is set (muon detection is done before hillas)
* number of pixels  passing tailcuts if >=3 (tailcut thresholds are
independent for muon and hillas)
* ellipsoid fit algorithm converges (i.e. no singular matrix)

dneise:
hmm  could it be something with ctapipe upstream again? .. did they change something and we suffer?

dneise:
hmm no idea at the moment.

digicampipe/io/containers.py
calispac: did we resolve this?

calispac:
:tada:

calispac:
So should we merge it now? :beers:

yrenier:
I just corrected the names. Once the tests are finished we can merge

dneise:
thanks @yrenier everything passed! I'm gonna merge it

Issue #253, opened 2018-10-16
1D and 2D histograming to be moved to cta-sst-1m/histogram?


calispac:
Currently we are using two histogram classes : 

1. https://github.com/cta-sst-1m/histogram (1D histogram and fitting)
2. https://github.com/cta-sst-1m/digicampipe/blob/master/digicampipe/utils/hist2d.py (2D histogram)

I think we should merge the two in the same repo (either the histogram repo or digicampipe)

To my opinion it should be merged within https://github.com/cta-sst-1m/histogram since these functionalities are not in principle useful for `digicampipe` only. 

What do you think?

moderski:
I was always wondering why we are using custom histogram classes, while numpy offers both .histogram and .histogram2d (even .histogramdd) for histogramming. I might be not fully aware of the full functionality of our solution, but is it possible to just use numpy?

moderski:
In such a case, I agree that hist2d.py should be merged within https://github.com/cta-sst-1m/histogram

dneise:
Actually we are using numpy histograms, at least in this implementation in `utils.hist2d.Histgram2D`... the class `Histogram2D` is merely a wrapper or a helper around the more ugly parts of filling a numpy histogram with multiple events.

https://github.com/cta-sst-1m/digicampipe/blob/master/digicampipe/utils/hist2d.py#L12

That is ... if I remember correctly... I totally omitted all comments which might explain what this class is good for :-(

yrenier:
It is used to make pulse template. Also it includes a Histogram2dChunked class which use some kind of buffer which makes the filling part much faster.

PR #255, opened 2018-10-16
Improve pulse template class & time reolution

yrenier:
Add standard deviation to the amplitude of the template.
Allow to create a template from data files.
Perform time resolution analysis as well.


digicampipe/tests/test_pulse_template_object.py
```diff
@@ -13,28 +13,54 @@
...
+    'digicampipe',
+    os.path.join(
+        'tests',
+        'resources',
+        'template_all.txt'
+    )
+)
+
+
+dir_data = '/mnt/baobab/sst1m/raw/2018/05/'
```
calispac: I suppose these should be removed

digicampipe/tests/test_pulse_template_object.py
```diff
@@ -13,28 +13,54 @@
...
-
     t, x = np.loadtxt(template_filename).T
-
     assert (t == template.time).all()
 
 
+def test_pulse_template_creation_with_file_with_std():
+    template = NormalizedPulseTemplate.load(template_filename_2)
+    t, x, std = np.loadtxt(template_filename_2).T
+    assert np.all(t == template.time)
```
calispac: does 
```python
assert np.all(template(time) == x) 
```
work too ?

calispac: ah it is not part of this function test actually

digicampipe/utils/pulse_template.py
```diff
@@ -2,41 +2,123 @@
...
+            return cls(amplitude=x, time=t)
+        elif len(data) == 3:
+            t, x, dx = data
+            return cls(amplitude=x, time=t, amplitude_std=dx)
 
-        t, x = np.loadtxt(filename).T
-
-        return cls(amplitude=x, time=t)
+    @classmethod
+    def create_from_datafiles(
```
calispac: To my opinion this includes to much stuff that are not related to a template building. And it repeats code that already exists in scripts/pulse_shape.py

This function should only be responsible to "fill_from" time and adc values the rest should be performed outside this class.

digicampipe/utils/pulse_template.py
```diff
@@ -2,41 +2,123 @@
...
+            cls, input_files, pixels=(0,), time_range_ns=(-10, 40),
+            amplitude_range=(-.1, 0.4), n_bin=100
+    ):
+        events = calibration_event_stream(input_files)
+        events = fill_digicam_baseline(events)
+        if "SST1M_01_201805" in input_files[0]:  # fix data in May
```
calispac: This should not be performed here! It does only benefit from the use of this function. 
We should solve this problem at the reading of the file in `zfits.py`

digicampipe/utils/pulse_template.py
yrenier: There are a few steps here:
- get pulse temporal position
- make a 2d histogram for each pixel of the normalized amplitude function of the relative time of the pulse
- fit each 2d histogram to get the mean and std charge for each time bin and each pixel
- combine all the pixels to get the template and uncertainty.
Which ones does not belong here ?

digicampipe/utils/pulse_template.py
```diff
@@ -2,41 +2,123 @@
...
-
-        if not (dt == dt[0]).all():
-            raise ValueError('time argument should be of constant sampling')
 
+class NormalizedPulseTemplate:
+    def __init__(self, amplitude, time, amplitude_std=None):
         self.time = time
         self.amplitude = amplitude
+        if amplitude_std is not None:
+            assert len(amplitude_std) == len(amplitude)
```
calispac: would be better to compare shapes since in principle this class works for 2 arrays as well
```suggestion
            assert amplitude_std.shape == amplitude.shape
```

digicampipe/tests/test_pulse_template_object.py
yrenier: as you wanted to start from the histogram, I replaced it by the histogram in 2fab9c5

digicampipe/utils/pulse_template.py
yrenier: I updated pulse_shape.py and changed create_from_datafile() to start from the 2D histogram in 2fab9c5

yrenier:
@calispac I tried to implement the change you requested. Any more comment ?

yrenier:
I also have the histograms for the 400 and 450 DAC levels, should I had them in the resources ?

digicampipe/scripts/pulse_shape.py
```diff
@@ -1,78 +1,121 @@
...
+        os.remove(output_hist)
+    charge_min = np.min(charge_range)
+    charge_max = np.max(charge_range)
+    integration_min = np.min(integration_range)
+    integration_max = np.max(integration_range)
     events = calibration_event_stream(input_files)
-    Rough_factor_between_single_pe_amplitude_and_integral = 21 / 5.8
+    events = fill_digicam_baseline(events)
+    if "SST1M_01_201805" in input_files[0]:  # fix data in May
+        print("WARNING: correction of the baselines applied.")
```
calispac: maybe we could put a "real" warning in the reader not necessary in this PR

digicampipe/scripts/pulse_shape.py
```diff
@@ -1,78 +1,121 @@
...
+        events = correct_wrong_baseline(events)
+    events = subtract_baseline(events)
     histo = None
-
-    for e in tqdm(events):
+    n_sample = 0
```
calispac: I see that histo, n_sample and n_pixels are declared for nothing right ?

calispac: forget what I said about histo...

digicampipe/scripts/pulse_shape.py
```diff
@@ -1,78 +1,121 @@
...
-    dset.attrs['extent'] = histo.extent
+    histo.save(output_hist)
+    print('2D histogram of pulse shape for all pixel saved as', output_hist)
 
 
 def entry():
     args = docopt(__doc__)
-    if args['--output'] is None:
-        args['--output'] = args['<input_files>'][0] + '.h5'
+    inputs = args['<input_files>']
```
calispac: I think at some point we should come with a class of something to handle command line arguments for all kind of scripts. 
We should also try to "harmonize" options / arguments names

digicampipe/utils/hist2d.py
```diff
@@ -16,14 +27,115 @@ def fill(self, x, y):
                 range=self.range
             )
             self.histo[pixel_id] += H.astype('u2')
+        self.xedges, self.yedges = xedges, yedges
 
     def contents(self):
         return self.histo
 
+    def fit_y(self, min_entries=2):
```
calispac: I would not call this fit_y since it is not really a "fit" but computation of mean and standard deviation along a coordinate axis.
Some we can maybe have mean() and std() methods

calispac: maybe when we move this to cta-sst-1m/histogram (see #253 )

digicampipe/utils/pulse_template.py
```diff
@@ -78,9 +100,8 @@ def plot(self, axes=None, **kwargs):
         t = np.linspace(self.time.min(), self.time.max(),
                         num=len(self.time) * 100)
 
-        axes.plot(self.time, self.amplitude, label='Template data-points',
-                  **kwargs)
-        axes.plot(t, self(t), label='Interpolated template')
+        axes.errorbar(self.time, self.amplitude, self.amplitude_std,
```
calispac: does it work if self.amplitude_std is None ?

digicampipe/scripts/pulse_shape.py
yrenier: There are data in May which are not affected, so that simple check would produce bad result if put in the reader.
The version of digicam/camera server was not yet present in that time in the header of the zfits, so we cannot use that.
If I knew how to do it in a proper way, I would indeed have moved it to the reader.

digicampipe/scripts/pulse_shape.py
yrenier: n_sample and n_pixel are declared outside of the for so their value are kept and can be used in the assert.

digicampipe/scripts/pulse_shape.py
yrenier: good idea, I opened an issue #261

digicampipe/utils/pulse_template.py
yrenier: old_template_plot
I just tried with the old template file (without std) and it worked as it is (no errorbar shown)

calispac:
Should we merge this beast together with the branch charge resolution when I have made some cleanup and tests ? Or you would prefer to merge to master directly ?

yrenier:
It was indeed quite big event before having to merge several PR from the master.
We probably don't need to add the charge resolution to make it even bigger, so I would rather merge it right away if you don't have any more comments/requests.

PR #259, opened 2018-10-22
compute PDE, XT, Gain drop as function of baseline shift

calispac:
This implements latest Baseline shift to PDE, XT, Gain drop function.
To be used in the `pipeline.py` at some point

### DIFF
--- a/digicampipe/calib/baseline.py
+++ b/digicampipe/calib/baseline.py
@@ -91,6 +91,42 @@ def compute_gain_drop(events, bias_resistance, cell_capacitance):
         yield event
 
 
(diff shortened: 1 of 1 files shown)


AndriiNagai:
PDE drop is true only for AD/DC scan, because it was calculated for a given wavelength of 468 nm. For real observation, PDE drop should be slightly different. 
The average PDE drop for NSB and Cherenkov spectrums will come soon...

calispac:
See this line : https://github.com/cta-sst-1m/digicampipe/blob/ba9839ec964e13ebd8352b0fbcff6563a9e1032a/digicampipe/calib/baseline.py#L127
At the moment it is just commented.

calispac:
I think we should have the NSB and Cherenkov spectra in digicampipe somewhere at some point

PR #260, opened 2018-10-23
use latest protozfitsreader version

calispac:


calispac:
I am wondering if github provides a link to download the latest release of a repo.
So far I found this https://github.com/cta-sst-1m/protozfitsreader/releases/latest

but this does not work:

```
pip install https://github.com/cta-sst-1m/protozfitsreader/releases/latest
```

dneise:
> I am wondering if github provides a link to download the latest release of a repo.

I wondered the same in the past, but then I realized that I rather want my dependencies fixed ... otherwise I would install new versions just by installing something later.

However, if you always want bleeding edge versions, I think you could do:

    pip install git+https://github.com/cta-sst-1m/protozfitsreader

which would install the master. Please not the protocol **`git+https`** at the beginning.

Issue #249, opened 2018-10-10
More test files required.


calispac:
Some of the integration tests require specific files. We should produce a set of these files for testing.
Here I am listing what I have in mind :

- [x] Dark data 100 events
- [x] Science data 100 events
- [x] NSB data : 0 MHz (no dark count), 40 MHz, 125 MHz, 600 MHz, 1 GHz
- [x] Dark data with pulsed light : 0, 1, 10, 100 p.e.
- [x] Zfits files with 16bits precision !!! (the tests files are still with 12bits precision)  
- [x] Hessio files from sim_telarray 

What do you propose?

calispac:
I will produce some NSB data with the toy Monte- Carlo

calispac:
I see that the file ``

looks like this:

![image](https://user-images.githubusercontent.com/23462345/46739845-bbf50c80-cca1-11e8-9f05-28339306ce0b.png)

We should definitely update it...

calispac:
For `zfits` file we can use the following first 100 / 200 events of these files : 

1. /sst1m/raw/2018/09/01/SST1M_01/SST1M_01_20180901_150.fits.fz (Science data, probably would need more than 100 evts in order to have a few "big" showers)
2. /sst1m/raw/2018/08/20/SST1M_01/SST1M_01_20180820_002.fits.fz (Dark data)

calispac:
@jurysek Could you provide some small sim_telarray files for which we could try to test and fix the `hessio` reader ?

calispac:
I have added a 55MB zfits file for the moment since we don't have smaller one that fits our needs

PR #262, opened 2018-10-30
More ressources

calispac:
See #249


yrenier:
Great ! before merging, it would be nice to have them used in the tests.
Also for reference, what shift are those files taken from ?

yrenier:
for the slow data, there is an easy way to extract a subset:
fitscopy 'DigicamSlowControl_20180901_000.fits['logs'][(timestamp<(1535837941944+1e4)) && (timestamp>(1535837941944-1e4))]' DigicamSlowControl_20180901_000_reduce.fits

PR #267, opened 2018-11-09
WIP: better pipeline

yrenier:
1 - show that the --pixel argument of the pipeline is not working and fix it
2 - treat bad pixels correctly taking average of sample of neighbor(FACT strategy)

### DIFF
--- a/digicampipe/tests/test_pipeline.py
+++ b/digicampipe/tests/test_pipeline.py
@@ -81,5 +81,40 @@ def test_pipeline():
             assert len(hdul[1].data[col]) == nevent
 
 
+def test_pipeline_two_pixels():
+    with tempfile.TemporaryDirectory() as tmpdirname:
+        dark_filename = os.path.join(tmpdirname, 'dark.pk')
+        hillas_filename = os.path.join(tmpdirname, 'hillas.fits')
+        compute_raw(
+            files=[example_file1_path],
+            max_events=None,
+            pixel_id=[0, 1],
+            filename=dark_filename
+        )
+        main_pipeline(
+            files=[example_file2_path],
+            max_events=None,
+            dark_filename=dark_filename,
+            pixel_ids=[0, 1],
+            shift=0,
+            integral_width=7,
+            debug=False,
+            hillas_filename=hillas_filename,
+            template_filename=template_filename,
+            parameters_filename=calibration_filename,
+            compute=True,
+            display=False,
+            picture_threshold=1,  # unusual value, so events pass cuts
+            boundary_threshold=1,  # unusual value, so events pass cuts
+        )
+        hdul = fits.open(os.path.join(tmpdirname, 'hillas.fits'))
+        cols = [c.name for c in hdul[1].columns]
+        nevent = len(hdul[1].data['local_time'])
+        assert nevent > 0
+        for col in expected_columns:
+            assert col in cols
+            assert len(hdul[1].data[col]) == nevent
+
+
 if __name__ == '__main__':
-    test_pipeline()
+    test_pipeline_two_pixels()


calispac:
Can we merge this with #262  ? And use the tests files ?

yrenier:
#262 seems unrelated to that one to me, but I don't really mind.
@calispac more related to that PR would be to use the charge determination used for charge linearity study. Could you take care of that ?

calispac:
Ok I will do !

calispac:
I have included the changes concerning the pulse integration. At the moment the saturated pulses are only "correctly" integrated. the charge to p.e. factor is assumed to be linear for the whole range. I have to include these non-linearity handling in an other PR. 

I also added a field to tell if any pixel in the signal is saturated, we shall see if saturation occurs often in Krakow.

Tests should pass

yrenier:
Thanks @calispac , but running the pipeline on the test files gives all reconstructed_number_of_pe > 1e3. Something must be wrong.

calispac:
Should be fixed now

calispac:
This was merged with #271 by mistake sorry..

yrenier:
adding the pixels with negative baseline shift to the bad pixels fixed the uniformity of the center of gravity of the showers
shower_center_map

digicampipe/tests/test_pipeline.py
```diff
@@ -102,7 +102,10 @@ def test_pipeline():
         # all data where Hillas computation succeeded have some intensity.
         assert np.all(data.intensity[good_data] > 90)
         # no shower are close to the center in the test ressources.
-        assert np.all(data.r[good_data] > 50)
+        assert np.all((data.r[good_data] > 50) & (data.r[good_data] < 650))
+        # cog of showers are within the camera
+        assert np.all((data.x[good_data] < 500) & (data.x[good_data] > -500))
```
calispac: :+1:

digicampipe/tests/test_hillas.py
```diff
@@ -0,0 +1,99 @@
...
+            saturation_threshold=3000,
+            threshold_pulse=0.1,
+            disable_bar=True
+        )
+        data, _ = get_data_and_selection(hillas_filename)
+    return data
+
+
+def test_correct_alphas():
+    data = get_data()
```
calispac: Could you also check if `correct_hillas()` + `compute_alpha()` from `digicampipe.image.hillas` give the same output ?

calispac: Then we should probably remove the variants

calispac: I personally think the two functions are much simpler since actually alpha is just |phi - psi| modulo pi/2. But I am not totally sure it is computing exactly the same thing.

(psi being the orientation of the shower longer axis, and phi the polar angle in the xy plane of the shower centroid)

digicampipe/image/hillas.py
```diff
@@ -108,10 +108,29 @@ def correct_alpha_3(data, source_x=0, source_y=0):
...
-    for i in range(len(alpha_cetienne)):
-        if (alpha_c_2[i] < alpha_c_1[i]):
-            alpha_cetienne[i] = alpha_c_2[i]
-    data['alpha'] = 180.0 / np.pi * alpha_cetienne
-    data['r'] = to_c_norm
-    data['miss'] = data['r'] * np.sin(data['alpha'])
+    alpha_cetienne[alpha_c_2 < alpha_c_1] = alpha_c_2[alpha_c_2 < alpha_c_1]
+    data.loc[:, 'alpha'] = 180.0 / np.pi * alpha_cetienne
+    data.loc[:, 'r'] = to_c_norm
+    data.loc[:, 'miss'] = data['r'] * np.sin(data['alpha'])
```
calispac: `miss` is probably wrong since `alpha` is now in degree right ?

calispac:
I see that miss is also wrong here : 
- https://github.com/cta-sst-1m/digicampipe/blob/f87f087b5f2d028eeb02d5bd59d97bcb56bbcb5f/digicampipe/image/hillas.py#L54
- https://github.com/cta-sst-1m/digicampipe/blob/f87f087b5f2d028eeb02d5bd59d97bcb56bbcb5f/digicampipe/image/hillas.py#L114

calispac:
See changes in https://github.com/cta-sst-1m/digicampipe/pull/274

digicampipe/scripts/follow_stars.py
```diff
@@ -0,0 +1,245 @@
...
+
+
+def nsb_rate(
+        files, aux_basepath, dark_histo_file, param_file, template_filename,
+        plot="show", plot_nsb_range=None, norm="log", disable_bar=False,
+        max_events=None, stars=('Capella',), mm_per_deg=-100.,
+        site=(50.049683 * u.deg, 19.944544 * u.deg, 209 * u.m),  # krakow
+        bias_resistance=1e4 * u.Ohm, cell_capacitance=5e-14 * u.Farad
+):
+    krakow = EarthLocation(lat=site[0], lon=site[1], height=site[2])
```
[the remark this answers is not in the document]
calispac: Actually could be the same for the catalog of stars

calispac:
Can we merge this soon ? this PR starts to become very big and we have a second "huge" PR as well #264

yrenier:
@calispac : Ok, I will not have time to finish it today, nor tomorrow. Everything should be tested, tests are passing, so we can merge if you want.

digicampipe/scripts/follow_stars.py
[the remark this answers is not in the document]
yrenier: I used the same coordinates than int the digicam-extra package.

PR #273, opened 2018-11-22
fix baselines in toy (to fix #265)

### LINKED ISSUE #265: Problem on Baseline in mpe script with toy datas
If we turn script fmpe on toy datas, we have problem with fill_digicam_baseline and substract baseline functions. So we need to comment thoses lines in mpe.py.

yrenier:
Small fix which allows the usage of baseline from toy data.
A bit optimistic in the sense that the baseline will always be the exact one, but it should be much better than NaNs.


calispac:
This baseline that is red from the file is the fadc offset i.e. it will always be the same depending on NSB

calispac:
I would rather prefer that we use baseline = 0 or compute "true baseline shift"

yrenier:
ah ok, that was too easy :-/

calispac:
I can add "the true baseline shift" in the toy data as well.

yrenier:
Adding the true baseline shift would fix the problem. 
As you are doing it, it may be nice to put the NSB as well.

calispac:
This will be added in : https://github.com/cta-sst-1m/digicamtoy/pull/21 
I will take this PR and change the reader of the toy data accordingly

calispac:
I also have issues now with event_id missing and time for MC data. Should I include a "now time stamp" and a "counter" event_id ?

calispac:
or maybe I just fill with 0

calispac:
Ok now time stamps and event ids are defined as the internal event number. 
Also some fields are only loaded once

calispac:
I believe this is ready to merge! Is it ok ?

yrenier:
looks goo to me

PR #274, opened 2018-11-26
Test hillas

calispac:
I am currently testing the 3 different ways we have to compute alpha and correct hillas parameters.

I have taken into a piece of paper the computation done and tried to use trigonometric properties to compute alpha more accurately and avoid InvalidValue warnings. 

I have added a few tests

Concerning method 2 and 3 they seem to fail sometimes to give accurate results but overall it seems fine


moderski:
If you need a reference to Hillas formulas I recommend http://adsabs.harvard.edu/abs/1993ApJ...404..206R

calispac:
Thanks @moderski  I just added it

PR #277, opened 2019-09-11
 calibration paper plots

yrenier:
Adapt scripts to produce the pulse template and timing resolution plots for the calibration paper.

### DIFF
--- a/digicampipe/utils/pulse_template.py
+++ b/digicampipe/utils/pulse_template.py
@@ -135,7 +135,7 @@ class NormalizedPulseTemplate:
             axes = fig.add_subplot(111)
         t = np.linspace(self.time.min(), self.time.max(),
                         num=len(self.time) * 100)
-        axes.errorbar(self.time, self.amplitude, self.amplitude_std,
+        axes.errorbar(self.time, self.amplitude, self.amplitude_std/2,
                       label=label, **kwargs)
         axes.set_xlabel('time [ns]')
         axes.set_ylabel('normalised amplitude [a.u.]')
--- a/digicampipe/visualization/time_resolution_plot.py
+++ b/digicampipe/visualization/time_resolution_plot.py
@@ -34,10 +34,7 @@ Options:
                                 resolution between any combination of 2 pixels.
                                 [default: 5.5]
 
-examples:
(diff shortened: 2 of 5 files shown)


digicampipe/utils/pulse_template.py
```diff
@@ -135,7 +135,7 @@ def plot(self, axes=None, label='Template data-points', **kwargs):
             axes = fig.add_subplot(111)
         t = np.linspace(self.time.min(), self.time.max(),
                         num=len(self.time) * 100)
-        axes.errorbar(self.time, self.amplitude, self.amplitude_std,
+        axes.errorbar(self.time, self.amplitude, self.amplitude_std/2,
```
calispac: Good catch ! :+1:

digicampipe/visualization/time_resolution_plot.py
```diff
@@ -154,13 +157,13 @@ def plot_resol(data_file, legend, ax=None):
     mean_charge_all, std_charge_all, mean_t_all, std_t_all, true_pe = \
         load_data(data_file)
     if ax is None:
-        fig, ax = plt.subplots(1, 1, figsize=(8, 6), dpi=100)
+        fig, ax = plt.subplots(1, 1, figsize=(8, 8), dpi=100)
     # ax.plot([20*pde, 2e3*pde], [1, 1], 'r-', label='requirement B-TEL-1380')
     # ax.plot([20*pde, 20*pde], [.9, 1.1], 'r-', label=None)
     ax.plot([20*pde, 2e3*pde], [3, 3], 'b--', label='requirement B-TEL-1640')
```
calispac: I think ax.axhline(3) would be better to use. same for the following plots

digicampipe/visualization/time_resolution_plot.py
```diff
@@ -154,13 +157,13 @@ def plot_resol(data_file, legend, ax=None):
...
-    ax.plot([20*pde, 20*pde], [1.4, 1.6], 'b-', label=None)
+    ax.plot([20*pde, 20*pde], [2.8, 3.2], 'b-', label=None)
```
calispac: what is this ax.plot, plotting actually ? I am a bit confused

digicampipe/visualization/time_resolution_plot.py
calispac: only annoying thing with ax.axhline() is that the xmin and xmax have to be given in relative coordinates (0 to 1)...

digicampipe/visualization/time_resolution_plot.py
yrenier: it is plotting the small vertical line at the beginning of the lines for requirement (to indicate that the requirement stops there)

